PR #16, opened 2019-02-18
V0.0 series

TeriForey:
Removed old tests, merged variants and updated tests.


TeriForey:
@PeteCausey-Freeman Can you have a look over this and check you're 100% happy with all the changes and that nothing else needs adding at this point.

PeteCausey-Freeman:
This seems to cover everything that we discussed. If you are happy with the testing then merge and tag. Thanks

Issue #44, opened 2019-05-15
Protein output incorrect for NM_005228.4:c.2237_2255delinsT


PeteCausey-Freeman:
[...]

Issue #41, opened 2019-04-30
REST API returns a warning using 'hg38' as genome build


beboche:
The REST API returns the following warning:

"Invalid genome build has been specified. Automap has selected the default build (GRCh38)"
When using 'hg38' as genome build like in:
https://rest.variantvalidator.org:443/variantvalidator/hg38/chr14%3Ag.88839501A%3EG/all

PeteCausey-Freeman:
Hi @beboche beboche
Agreed. The same is not apparent for the hg19 equivalent so this is a bug
https://rest.variantvalidator.org:443/variantvalidator/hg19/chr14:g.89305845A>G/all

For now please use GRCh38 instead of hg38 while this is sorted.

PeteCausey-Freeman:
Closed by fix #45

PR #45, opened 2019-05-15
Bug fixes for #44 and #41

PeteCausey-Freeman:
VariantValidator now correctly handles delins where the variant protein is shorter than the reference protein


TeriForey:
Code was extracted and merged via PR 47

PR #74, opened 2019-07-11
Fix tx update

PeteCausey-Freeman:


PeteCausey-Freeman:
This should be sufficient for a final v0.0 series tag. Please add an appropriate version number, merge into master and tag.

TeriForey:
Thanks Pete, it looks good! I need to finish what I'm doing with v1 but then I'll get right on this.

PR #78, opened 2019-07-26
Vf integration

PeteCausey-Freeman:
There were a couple of bugs that I spotted in the code and the test files around the use of hp.parse(). It should be hp.parse_hgvs_variant()

I have also added a few normalizer instances. These will phase out existing instances as we begin to integrate Ensembl data

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ For sequence variations falling within the open reading frames of genes, Variant
 
 ## Pre-requisites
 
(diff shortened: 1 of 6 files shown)


PeteCausey-Freeman:
Note, I could not get one of the tests to work because I was not entirely sure how you want to handle it. It's an operational test listed in the commits.

PR #90, opened 2019-09-19
Develop v1.0

PeteCausey-Freeman:


PeteCausey-Freeman:
Fixes travis and provides an updated VV dataset

PR #260, opened 2021-01-15
Added single letter code functionality

AliJ752:
Hey Naomi, Ive added some single letter functionality to the input checks (may not need it if we take variant validator just in case)


Istaisa:
Looks good

PR #261, opened 2021-01-18
Added format check for string input

Istaisa:
Added a check for HVGS nomenclature for the string - currently does exceptions through raise SystemExit but open to discussion!


AliJ752 (approved):
Looks good to me!

PR #293, opened 2021-08-17
Update to Travis.com

PeteCausey-Freeman:


ifokkema:
A word of warning.... Despite them saying it is, Travis-ci.com is not free! It gives you credits, that run out pretty fast. Better go and change your settings to have tests run only at PRs or so (so not every push triggers a build). From the top of my head, you only get some 1000 minutes for free. How fast that runs out depends on your tests of course. Mine take some 15 minutes, so including the booting of the machine and the setup, we're looking at some 18 minutes per build. So I'm only getting ~55 runs for free. Not sure if the credits reload at the end of the year or if these are one-time credits. You can buy 2500 minutes for 15 USD I think, so it's not too bad, but definitely not free anymore.

PeteCausey-Freeman:
Nice one. Thanks. Are the settings controlled within the travis website?

PeteCausey-Freeman:
Found it. Thanks for the tip

ifokkema:
> Found it. Thanks for the tip

Cool, no worries!

So on Linux, 10 credits are 1 minute build time. I'm checking how fast I'm running out, and will likely pay 15 USD per year to keep running on Travis as that's cheaper than trying to figure out how to move the tests elsewhere (I'm not cheap lol).

PeteCausey-Freeman:
Agreed. 10,000 credits a month should do us and if necessary I don't mind paying an extra 15 quid occasionally from my own pocket. I'll bill the universities with 1000 % inflation ;)

ifokkema:
Haha sounds profitable :rofl: 
Note that the credits don't reload every month. Not sure if they reload at all... so it might just be 10,000 credits one-off.

PeteCausey-Freeman:
Hmm. They don't make that clear on the website either. Will have to see how we get on. Thanks again

ifokkema:
Yeah, it's super unprofessional. When I googled for more info, all I found was loads of rants on the subject, but very little info. I'm glad I switched late as you did, so I could enjoy a few more months of free builds :sweat_smile:

PR #345, opened 2022-02-16
Tandem repeats dev rob

RSWilson1:
Merging Rob and Becky's changes to final version.


RSWilson1 (commented):
Checked with current Expaned_repeats version and all up-to-date.

PR #348, opened 2022-02-16
Tandem repeats dev rob

RSWilson1:


RSWilson1:
added test -- coverage at 80%

PR #563, opened 2023-12-12
Improve how SeqRepo errors appear to users

John-F-Wagstaff:
Hello Pete this looks like the right place to put the improved error message but I ended up being a bit verbose, can you have a check and see what you think, I will update as necessary.


John-F-Wagstaff:
Improved wording, review (in meeting) from Pete.

Peter-J-Freeman (commented):
Looks like the correct place to me. We will monitor to see if there are any other places it is needed

PR #635, opened 2024-07-18
Update transcript version warnings to include genome build info for a…

Peter-J-Freeman:
…lignments available


ifokkema:
Can we set up some kind of system that notifies me when warnings or error messages change? Or, is there some kind of index or dictionary where I can see what possible warnings or error messages can be created by VV? Since there are no error codes, LOVD parses the strings sent by VV to try and figure out what is information, a warning, or an error. When the strings change, LOVD doesn't know what the message means anymore since it doesn't match any of the pre-configured strings. I noticed yesterday during the VIJ meeting (after we discussed providing feedback about the availability of new transcript versions) that my tool now considers using an older transcript version as a fatal error because the string changed somewhere along the way. I sometimes add new strings when I notice a new message I don't match yet, although this probably also means some other message can be removed. However, I don't have an overview of when things change or what possible errors exist. Of course, having permanent error codes would be even better, but as an alternative, can we set up some system that notifies me so I can adapt the LOVD-VV library in time?

Peter-J-Freeman:
We definately need some discussion around this. Some serious issues have arisen since we migrated to the Ensembl data.

I think ErrorCodes are the way to go. We did start including these. In this case we could add the prefix TranscriptVersionError: (or warning perhaps TranscriptVersionWarning: )

We need to go through the code base and add error prefixes to all error messages and document them. This is a big job, but I should get on with it over the summer!!!!

Peter-J-Freeman:
Note, this error message will stay. Do you want me to add the prefix and push a new version @ifokkema. If so, what prefix?

Peter-J-Freeman:
Have added in the flag TranscriptVersionWarning: as the error code here

Peter-J-Freeman:
Opened an issue @ifokkema

PR #642, opened 2024-08-09
Updtes to the expanded repeat code for simple repeats only so far

Peter-J-Freeman:


Peter-J-Freeman:
@John-F-Wagstaff

We need to add in tests for g. NC_ LRG and NG_ but this is what I have so far. Please take a look, test and see what you think

This is nice because it pulls in work from STP students

John-F-Wagstaff:
So far so good. I need to take some time to do the testing but the initial read-through of the main commit looks good.

However there are a few leftovers in parse_repeat_variant from it's experimental state, some of which we might want to change before we start using them. 
- Firstly we have a few efficency issues. 
  - We use ' re.IGNORECASE' a number of times IGNORECASE is as far as I know always slower than specifying a pair of characters (it has to handle region specific painful stuff, what capitals are is not apparently consistent between locales) micro-benchmarking shows  '[gGaAtTcCuU]+$'  takes ~9.5 seconds to do an amount of work that '[gatcu]+$' with re.IGNORECASE does in ~14.5, making the IGNORECASE version ~50% slower. Other bits from this code I tested go from about 40% to 50% longer in the current form than the long version without IGNORECASE.  This appears to be the normal performance impact, so we probably want to fix this whenever we make changes to a region.
  - Putting a limit into split,or re.split, gives a small speed boost for short variants but this grows with the variant length, halving time taken, or more for longer variants!  so wherever we can we want split(':',1) or equivalent.
  - Using partition which I have only just myself found out about is even better, when we can cope with its limitations without extra code (that is that partition is text only, no regex, and always returns without error, even if split value is missing)
- Secondly you have left in some of the assert catch re-raise logic that should not be this complex.  We might want to simplify if we can, though you already removed the worst of it so this is probably just me being nit-picky.
-  Thirdly we have some direct self.hp.parse usage. We do want to remove this kind of variant creation, but we may want to leave this to a later commit, maybe even write some helper functions when we do?

Peter-J-Freeman:
@John-F-Wagstaff Yep, I forgot to mention that I was passing on to you as the first exercise in direct HGVS object creation rather than parsing. One for you to investigate. For the other improvements, can you tweak these once you get chance. Great catches. Think we need to start going through the entire code base to remove re-parses ASAP.

John-F-Wagstaff:
OK I have run the tests and a bit of my own testing too, so far so good.

@Peter-J-Freeman  Does this mean that you want me to contribute to this branch/pull request? i.e. not just review but add too? I can if you want, but I need to check, trying both to push to the same branch at the same time is a headache.

If you do want me to make changes then I am probably thinking of
1. Add tests for NC_ as well as LRG and RSG (NC_ first)
2. Add simple optimisations (case issues and add limit at least)
3. Attempt direct hgvs variant creation (which may warrant a helper function that might technically be better organised if placed in a different pull request? ) 

Does this seem OK? 
We may be better off for the direct hgvs creation pushing as-is, without the direct creation, and then making a distinct feature branch for just that. Too many things in the same branch causes issues, but I can push up-to the end of 2 then we can discuss, depending on what it ends up looking like.

Peter-J-Freeman:
@John-F-Wagstaff . Happy to do whichever way works best. Yes, please make changes and add some tests. Then we can trouble shoot together. Happy for you to do the best way for you

John-F-Wagstaff:
Hello @Peter-J-Freeman
This should all **finally** work.  There was a bit more to do than I thought going in but it should be a lot better now.
We might be better off adding a revised data to the credits at the top of the file, but for now it should all be OK. Would you be able to have a quick look? Even if we don't make any changes to this patch set feedback on style issues etc would be good for any later additions I end up making to VariantValidator.

PR #701, opened 2025-03-24
Develop pf 2025

Peter-J-Freeman:
Hi @John-F-Wagstaff @leicray @ifokkema 

John, please can you take a look at this rendition. A 1st attempt at the LOVD synatax chacker. 

Uses Cli interface and falls back to web if needed

To setup you need to need to run

```bash
pip install -e .
python -m VariantValidator.bin.setup_lovd_syntax_checker
```

Logic, this installs the LOVD data straight into the VV app. Each time we update the app, we can run the setup script. 

Once setup  run some dodgy variants like 

```python
import json
import VariantValidator
vval = VariantValidator.Validator()
variant = 'NM_014712.2:exon17:c.G>C'
genome_build = 'GRCh38'
liftover_level = False
lovd_syntax_check = True
select_transcripts = 'mane'
validate = vval.validate(variant, genome_build, select_transcripts, liftover_level=liftover_level, lovd_syntax_check=lovd_syntax_check)
validation = validate.format_as_dict(with_meta=True)
print(json.dumps(validation, sort_keys=True, indent=4, separators=(',', ': ')))
```

Note, in the bug catcher, I added a note. We perhaps are raising an exception where we should be continuing. This is likely why all batch jobs which have 1 bad variant fail rather than trying to process remaining variants. We may be resposible for some of @leicray 's hair loss over the last few years by not spotting it. Can we make this code compile an unhandled error rather than a raise?????


Peter-J-Freeman:
@John-F-Wagstaff 

In addition, can you review this variant

```python
import json
import VariantValidator
vval = VariantValidator.Validator()
variant = 'NM_005159.5 (ACTC1): c.809-58_809-13delinsTG[14]'
genome_build = 'GRCh38'
liftover_level = True
select_transcripts = 'all'
validate = vval.validate(variant, genome_build, select_transcripts, liftover_level=liftover_level)
validation = validate.format_as_dict(with_meta=True)
print(json.dumps(validation, sort_keys=True, indent=4, separators=(',', ': ')))
```

Once fixed, we need to remove print statements and hash out the import traceback and traceback print from line 662 of format_converters.py

John-F-Wagstaff:
Hello Pete

I have had a look and fixed my error catching patches to work with the added LOVD code, rather than leading to the LOVD code getting skipped. The pull request to pull this onto your work is https://github.com/openvar/variantValidator/pull/702 Which you might want to use, otherwise we could split the patches up differently (e.g `git cherry-pick 28c119de0f8b98eab4c339aa7c9412c4addb1a24` to put the first into this branch, and then the same for the second 2 commits, with  fix_develop checked out, to put them into into fix_develop) or else do other changes if you wish.

For the bug catcher this is to catch unexpected internal failures, not the expected type of error. As such we probably want to, at most, make this optional, and we probably need to make it log somewhere if we do make it skip-able. Many users do not complain, so the only time we get to find out about some errors is when we get a crash, worse normal website users and batch uses seem to submit different types of failing variant, so we need to find out when this triggers somehow. If we switched the website to sending out nightly emails to the admin address, when there are new failures since the last evening, or something like that, then that would be a lot safer for us in the long term than just skipping. Still so long as we get these unexpected failure types logged somewhere, and somehow set the server up to notify us about them then I see no reason not to make skipping past broken variables possible. It would unfortunately be very easy to push in a fix that shut it up silently (i.e. raise->continue) but getting notification right is much harder.

Peter-J-Freeman:
@John-F-Wagstaff . Actuallym yes, this is exactly why we left it as raise. To make sure we didnn't miss bugs. That's a good reason to keep the raise indeed. One to think about. 

If you want to split the patches to make it flow better, please do so and complete the merges. Let me know what to pull.

I will keep sorting the indel with many spaces and a gene symbol that @leicray sent us. Realised what may be going wrong.

John-F-Wagstaff:
@Peter-J-Freeman if we are intending to trigger this pull request quite quickly then it should be fine just pull https://github.com/openvar/variantValidator/pull/702

If you are intending to keep working on this pull request for a while we might want to do something more complex, like splitting the set of patches out into 2 pull requests. This would allow us to reduce crashes in the fix_develop branch, which might be useful if you are running that on our server, but that has issues due to patch order problems, basically it makes it flow worse in return for getting the fix in quick, so it depends, how soon is this going to be triggered? This was why I was asking, sorry for not being clear as to why. So if you are intending to finalise this pull request relatively quickly then just using https://github.com/openvar/variantValidator/pull/702 is the best thing to do.

Peter-J-Freeman:
Just need to fix one bug on this pr then will merge.

We need to increase the version number because it's a breaking change. Even the environment is changed. What do you recommend as a version number?

John-F-Wagstaff:
No Idea, normally I would say 4.0.0 but it is a bit soon after our last major release, so it should be at least 3.1.0 but which is a bit of a hard one.

Peter-J-Freeman:
Exactly my thoughts.

We could just bite the bullet and go for 4.0.0 with notes about why we made this decision. It's a majorly step up.

Vvhgvs needed an update. Vv too. The whole environment changed. I think we should consider 4.0.0

VariantValidator/bin/setup_lovd_syntax_checker.py
```diff
@@ -0,0 +1,52 @@
...
+LOVD_DIR = get_installation_path()
+WEB_DIR = os.path.join(LOVD_DIR, "web")
+
+# Define file paths
+PHP_SCRIPT = os.path.join(LOVD_DIR, "HGVS.php")
+AJAX_SCRIPT = os.path.join(WEB_DIR, "ajax.php")
+INDEX_SCRIPT = os.path.join(WEB_DIR, "index.php")
+
+# Define URLs
+HGVS_CHECKER_URL = "https://raw.githubusercontent.com/LOVDnl/HGVS-syntax-checker/main/HGVS.php"
```
ifokkema, line 26: Do we need to create versions so you can fixate the version of the LOVD/HGVS library to use? When I'm just adding functionality, it may not matter much, but in case I will change something, it would be good not to break anything on your side.

ifokkema: > Actually, I need to add code that gives the LOVD syntax checker version. Thanks for the pointer. Will add to the pr before merging

Currently, the library's "getVersions" call returns:
```json
[
    {
        "library_version": "2025-03-19",
        "HGVS_nomenclature_versions": {
            "input": {
                "minimum": "15.11",
                "maximum": "21.1.2"
            },
            "output": "21.1.2"
        }
    }
]
```

You can cheat and combine them in the CLI. The output of `php -f HGVS.php "c.100del" "getVersions"`:

```json
[
    {
        "input": "c.100del",
        "identified_as": "variant_DNA",
        "identified_as_formatted": "variant (DNA)",
        "valid": false,
        "messages": [],
        "warnings": [],
        "errors": {
            "EREFSEQMISSING": "This variant is missing a reference sequence."
        },
        "data": {
            "position_start": 100,
            "position_end": 100,
            "position_start_intron": 0,
            "position_end_intron": 0,
            "range": false,
            "type": "del"
        },
        "corrected_values": {
            "c.100del": 1
        }
    },
    {
        "library_version": "2025-03-19",
        "HGVS_nomenclature_versions": {
            "input": {
                "minimum": "15.11",
                "maximum": "21.1.2"
            },
            "output": "21.1.2"
        }
    }
]
```

The API returns these versions in the output, in the "versions" key.


> As for how you version, we take from your master branch and API, so if the general structure does not change I.e. You don't remove or change existing keys, things should not break.

So, if we ever do anything that would be backward-incompatible, I'll release API version 3, so as long as you use "v2" in the URL, you'll be OK (that said, if I would break something very soon, I'll just give you a heads-up and keep "v2"). However, the HGVS library currently only uses dates as versioning, but that's probably not a great idea now. So maybe I should create some tags and stick to semantic versioning, which will require me to update the APIs output (add the semantic version and keep the date as well, or just replace the date with the semantic version).


> We could do versioning in the future if necessary. Would pull the files from a specific tag at your end.

Hmm... I suddenly realize that tagging probably won't allow you to pull in, e.g., "1.x.x", right? Unless I create and constantly update a "1.x.x." tag. And sticking to a very specific version, e.g., "1.0.0" will be labor-intensive as I have lots of planned updates still laying around and getting each update into VV would require you to update the version to download and re-download the version... or just a manual re-download will also work... hmmm....

leicray:
> We could just bite the bullet and go for 4.0.0 with notes about why we made this decision.

A compromise that would still convey the notion of a major upgrade would be 3.5.0.

leicray:
For future reference, we do need to ensure that we never have version 6.6.6 😊

Peter-J-Freeman:
You know that I hadn't thought about that and now want to release it.

Imagine. V6.6.6 Integration of the lovd syntax checker. @ifokkema would be pleased. 😂

Peter-J-Freeman:
@ifokkema. I want to add the LOVD SC version. 

But which is the best to add from this?

```json
"versions": {
"library_version": "2025-03-21",
"HGVS_nomenclature_versions": {
"input": {
"minimum": "15.11",
"maximum": "21.1.2"
},
"output": "21.1.2"
}
```

Its the tool version we are after. For the ClI I need one of these to get the data right

```bash
php -f HGVS.php versions
php -f HGVS.php getVersions
```

ifokkema:
> Imagine. V6.6.6 Integration of the lovd syntax checker. @ifokkema would be pleased. 😂

:laughing: 
:smiling_imp: 


> @ifokkema. I want to add the LOVD SC version.
> 
> But which is the best to add from this?
> 
> ```json
> "versions": {
> "library_version": "2025-03-21",
> "HGVS_nomenclature_versions": {
> "input": {
> "minimum": "15.11",
> "maximum": "21.1.2"
> },
> "output": "21.1.2"
> }
> ```

The "library_version".  The other versions are HGVS nomenclature versions; what is supported for input (min, max), and what is the supported output. Adding HGVS nomenclature versions was recommended for tools so a user could easily see if the output was up to date with the latest nomenclature and whether it would support old or very new nomenclature as input.


> Its the tool version we are after. For the ClI I need one of these to get the data right
> 
> ```shell
> php -f HGVS.php versions
> php -f HGVS.php getVersions
> ```

Yes; see my comment above on how to get it out while using it, so you won't have to make a separate call.

VariantValidator/bin/setup_lovd_syntax_checker.py
Peter-J-Freeman: Updating from master isn't a problem
We just run the setup script.

What I'll do is setup some unit tests for the code that calls your api and cli. If your structure changes and breaks the tests, we will update the code.

Apparently we can pull thel files directly from a gitgub tag

wget https://raw.githubusercontent.com/OWNER/REPO/TAG/PATH/TO/File. 

Any chance you can make a 0.0.1 tag and label it as a dev or pre release  @ifokkema so we can test it?

ifokkema: I went back in time and added a bunch of tags where appropriate. It also means that we're currently at v0.4.1.
https://github.com/LOVDnl/HGVS-syntax-checker/releases/tag/v0.4.1

Since we're then using version numbers, it makes sense for me to add that to the output, which means the structure will change a bit. I'll then have a `library_version` (like now, but then no longer containing a date), and a `library_date` field (containing the date). That makes the most sense, I think. And that will then also be v0.4.2.

PR #726, opened 2025-05-01
Fixing up bugs

Peter-J-Freeman:


Peter-J-Freeman:
@John-F-Wagstaff If you have time can you make the edits. I will open the new issue for LOVD testing. I want to know why the server is pinging the API and not the CLI. It should not be

John-F-Wagstaff:
@Peter-J-Freeman What where you intending to happen, in lovd_syntax_check_g2t, when there is a error? If I am going to make changes to the indenting doing a return early on error and de-indent for this is also probably a good idea. Do we want to return the error as a message (and the corrections as empty), or return both the messages and corrections as empty?

Peter-J-Freeman:
I think that if we hit an error and we and LOVD cannot handle it, return the default error 

```pyton
return {'error': 'Unable to recognise HGNC ID. Please provide a gene symbol',
                                "requested_symbol": submitted}
```

So a try except is likely needed or someting else

John-F-Wagstaff:
Hello Pete, I had just finished running tests when you sent in cb38a9a but I had done some changes to the ORF handling bit that this conflicts.  I had in fact removed this section entirely, as the whole ORF part of the query gets cut off any standard transcript query before use, unless you don't have a version number for your transcript (in which case your query will fail), so this does not even change the case for the error message. The LOVD results for this where, and are, also not used.  

Do you want me to put handling in to add this LOVD output to the user returned results? do you also/instead want me to trim the ORF part off any inputs? or should I just cut this section of the code out?

Peter-J-Freeman:
Hi John, the ORF handling is needed for gene symbols like C3orf52. I am going to complete this merge, then we can look at what you have. I did make a correction to the LOVD handling in the ORF section already. Switched the LOVD on only if requested since it cause a bug if not in place

Peter-J-Freeman:
@John-F-Wagstaff I merged into develop. If you push up your code and PR into fixing_up_bugs, I can locate the changes and manually update.

John-F-Wagstaff:
@Peter-J-Freeman I just put in a pull request for a version tweaked to work against the current state of the code (at https://github.com/openvar/variantValidator/pull/728 )

PR #738, opened 2025-06-16
updates to the fallback onto LOVD error messages and warnings, see ht…

Peter-J-Freeman:
…tps://github.com/openvar/variantValidator/issues/737

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Declare the base image
-FROM python:3.11
+FROM python:3.12.11
 
 # Create the WorkDir
 WORKDIR /app
@@ -16,7 +16,8 @@ RUN apt update
 # Install apt managed sofware
 RUN apt -y install git \
     postgresql-client \
-    sqlite3
+    sqlite3 \
+    php
 
 # Upgrade pip
 RUN pip install --upgrade pip
--- a/VariantValidator/modules/valoutput.py
+++ b/VariantValidator/modules/valoutput.py
@@ -237,6 +237,23 @@ class ValOutput(object):
         if "lovd_api_error" not in variant.lovd_syntax_check.keys():
             lovd_messages = {}
             lovd_corrections = {}
+
+            try:
+                for key, val in variant.lovd_syntax_check['data'][0]['corrected_values'].items():
+                    lovd_syntax_suggestions = f"LovdSyntaxcheckSuggestions: [{key}, {round(val, 2)}]"
+                    if val == 1:
+                        variant.warnings = []
+                        if key == variant.original:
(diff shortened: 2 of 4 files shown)


John-F-Wagstaff:
Hello Pete

This looks OK, code wise but a few comments

Firstly, for the install requirements, we have set more to ==. This is not necessarily wrong but it is something we are going to need to keep an eye on, and have some kind of system/schedule to check and update.  If we don't then we could get locked to old, even archaic versions over time, particularly when the compatibility limits of one older component locks out newer versions of others, basic shared dependencies can be surprisingly incompatible between versions in python.

Secondly the Dockerfile updates look valid, but given how it is used I can't directly test without getting a local jenkins version set up, I tried using jenkinsfile-runner but it had version compatibility issues with the java libraries on my system so no luck on a more partial test. The code changes match the working version for the API though so it should be fine, but if you want me to run a local test I will need a bit more time.

Thirdly we might want to be able to prevent fallback to LOVD web.  Silent fallback to the web version when the local version is installed but not working could hide issues in the future. If we do then we may want to put LovdSyntaxcheckLibraryVersion (and or the base URL for the LOVD check) into our metadata, rather than putting it into the warnings. This sligtly expands the effects of this patch, and might be out of scope for this pull request, but it is something we might want to think about later. if so.

The last, possibly nit-picky issue is that you have merged patches for 2-3 unrelated issues into 1. It would be more neat to keep the install requirements and Docker work in a separate patch to the code changes, separating by function like this can make finding errors by git-bisecting or other such processes simpler and is generally encouraged even if it does not affect the actual changes suggested.

Peter-J-Freeman:
Hi @John-F-Wagstaff. The == sections were a vain attempt to lock into fast ans stable versions of some modules. I will remove these and re-test.

Don't worry about the docker install. It is rarely used and I think we may remove it and only allow the rest api to be built through docker since it makes more sense and does contain a complete VV build anyway. I can get it running locally.

> The last, possibly nit-picky issue is that you have merged patches for 2-3 unrelated issues into 1. It would be more neat to keep the install requirements and Docker work in a separate patch to the code changes, separating by function like this can make finding errors by git-bisecting or other such processes simpler and is generally encouraged even if it does not affect the actual changes suggested.

Agreed,  this was caused by lack of coffee, too much rushing and other distractions :)

> Thirdly we might want to be able to prevent fallback to LOVD web. Silent fallback to the web version when the local version is installed but not working could hide issues in the future. If we do then we may want to put LovdSyntaxcheckLibraryVersion (and or the base URL for the LOVD check) into our metadata, rather than putting it into the warnings. This sligtly expands the effects of this patch, and might be out of scope for this pull request, but it is something we might want to think about later, if so.

I will push up a new env build tested for now. Can you pull and add in this bit of code? will add a note here once its good to go

PR #742, opened 2025-07-22
Setup logging from config filepath

Peter-J-Freeman:


John-F-Wagstaff:
Hello Pete

We probably want to add the log file location to both branches(with and without ` config['logging'].getboolean('log') is True`), so that setting logging to true does not make the code ignore the log file location
so putting in something like

    if  config['logging']['file_name']:
        settings.LOGGING_CONFIG['handlers']['file']['filename'] = config['logging']['file_name']

at line 20 in  
VariantValidator/logger.py
(just where you deleted a line in the current patch), as well as this change, is probably necessary to consider this fixed. This is important as the logging True path is the one where we get to set the log level, which is likely something we will need if we start getting mysterious failures on the server.

Peter-J-Freeman:
@John-F-Wagstaff ok, pushed up ready for checking

Peter-J-Freeman:
@John-F-Wagstaff Tests all pass now. Allows backwards compatibility with old conf files. Also, I updated vv_hgvs to remove a dependency on pkg_resources

John-F-Wagstaff (approved):
Looks good now with the edits, should be a bit more robust too
Thanks

PR #749, opened 2025-08-18
Expanded repeat in out

Peter-J-Freeman:
@John-F-Wagstaff. Here is the code for expanded repeat syntax with a set of unit tests and updates to the existing functional tests. 

Please can you review the parsing. There are a few parse statements in the new script.

### DIFF
--- a/VariantValidator/modules/indel_to_expanded_repeat.py
+++ b/VariantValidator/modules/indel_to_expanded_repeat.py
@@ -76,9 +76,10 @@ def decipher_repeated_unit(sequence):
             return unit
     return sequence  # Default: no recognizable repeat pattern
 
-def decipher_start_of_full_reference_repeated_sequence(reference, repeated_unit, start, validator):
+def decipher_start_of_full_reference_repeated_sequence(reference, repeated_unit, start, validator, window_size=100):
     """
-    Given a position where a repeat starts, walk backwards to find the full beginning of the repeat block.
+    Given a position where a repeat starts, fetch a larger window of sequence
+    and walk backwards in-memory to find the full beginning of the repeat block.
     """
     if not repeated_unit:
         raise RepeatedUnitError("Repeated unit must be a non-empty string.")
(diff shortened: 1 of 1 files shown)


John-F-Wagstaff:
OK, I will have a go, hopefully it won't be too hard.  I will need a review, from you, on my changes after.

Peter-J-Freeman:
I uploaded a new push [Update indel to tandem repeat code to do 2 long fetches and not many …](https://github.com/openvar/variantValidator/pull/749/commits/c663e0beef195526a1fe92c0f33a8fc9b4f7353a)

It addresses the seqrepo fetching and also the use of a knoen tandem repeat block, with a fall back if it is not passed.

To do

- The code still has parses, i.e. works with strings. But the root code will always pass strings and not objects because it comes from the finalised output objects. So we will need to parse into an object within the script at several places, but we want to go straight to object rather than through the slow parser
- File and function name changes. I left these since you mentioned you wanted to look at the code etc to make sure it fits with your VRS work.

> PS: we have discussed other changes WRT the VRS work, specifically so that the outputs of indel_to_expanded_repeat.py are still valid hgvs objects, but it is probably best to push this first as a complete change then add the VRS work on top as a separate pull request. The alternative would be much more complicated and possibly require merging the VRS branch into this one, muddling the logic and also adding both more complexity and other delays.

I can update the code so it rerurns both the string of the expanded repeat and also parsed HGVS objects????? Perhaps in a dict?

John-F-Wagstaff:
> do you want the outputs e.g. NC_000004.11:g.3076606_3076662GCA[21] to be parsed HGVS objects. 


Not yet, they won't work at the moment. 

To get this to work we have to add a twiddle to the local hgvs object handling to take the repeat and do something sensible with it, which is probably best left as a Pre-patch to my VRS work, as opposed to adding it to this set. If you leave the output as text that should be fine for now.

John-F-Wagstaff:
> I can't think of any good names. 

There isn't really a good one, the best I could come up with is sequence state, shortened to seqstate/SeqState, but that is still a bit ugly to say the least. Still if you are already porting to a new branch/ creating a new set of patches (which could make re-naming the files less ugly), and don't have anything better either, then that would be my best shot.

Peter-J-Freeman:
seqstate/SeqState is fine. But since the file is about repetitive sequence how about repeatstate??? still ugly, more descriptive :)

John-F-Wagstaff:
We currently use the same logical pattern of naming repeatedly,
indel -> tandem repeat
we do this in `indel_to_expanded_repeat`, `convert_indel_to_expanded_repeat`, and the use of `indel_info` for the hgvs postion edit. 
This encodes the fact that input is devoid of repeat status understanding, but the output contains repeat information (but confuses the intent of the code, as it does not match = type variants, or even just del or just ins variants to a certain extent). 

We can substitute
seqstate -> tandum repeat
into that logic without having to rewrite much, or changing the rest of the naming and annotation logic, getting `seqstate_to_expanded_repeat`, `convert_seqstate_to_expanded_repeat`, and `seqstate_info`.  (or even `seq_state_info` etc. if you prefer)

We would have to change more wording to change it to repeatstate, as the fact that the input is devoid of repeat status understanding is part of the logic of the naming at the moment and so more has to be changed to make it make sense. Fore example `repeatstae_to_expanded_repeat` does not convey the annotation change as clearly,  nor does `convert_repeatstate_to_expanded_repeat`, they sound like the input should be repeat aware, and may not make sense in context.

If you are OK with more changes then this is not a deal breaker but it would have to be something like `derive_repeatstate` instead of the current naming logic, and there would need to be other changes to the comments/docstirngs to match this and make the logic clear to readers. Hence me choosing to suggest seqstate as the lazy option.

Peter-J-Freeman:
I say keep it simple. Go with your seqstate. If it works it works :).

PR #750, opened 2025-09-02
add configurable php path for LOVD tool

Peter-J-Freeman:
@John-F-Wagstaff This should allow us to configure PHP on the servers.

### DIFF
--- a/VariantValidator/bin/lovd_syntax_checker.py
+++ b/VariantValidator/bin/lovd_syntax_checker.py
@@ -1,9 +1,24 @@
-import os
 import subprocess
 import json
 import time
 import threading
 import importlib.resources as pkg_resources
+import os
+from configparser import ConfigParser
+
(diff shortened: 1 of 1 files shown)


VariantValidator/bin/lovd_syntax_checker.py
John-F-Wagstaff, line 1: why do
PHP_PATH = None
and then all the conditionals? 
wouldn't
PHP_PATH = 'php'
work reduce the complexity quite a bit?

John-F-Wagstaff: OK, Do you want to re-base and push -f to tidy up? 
You can do
`git rebase -i HEAD^^`
then select `f` or `fixup` instead of `pick` for the last patch (this merges the fixup patch into the previous patch without changing the commit message), then force push with git to update the branch. It is bad on shared branches but for small one user feature/fix branches like this is standard practice on some projects to avoid git clutter.

Issue #751, opened 2025-09-05
Improved warning message when no transcripts fully overlap the described variation


leicray:
**Is your feature request related to a problem? Please describe.**
Validation of some variant descriptions can generate the warning "No transcripts found that fully overlap the described variation in the genomic sequence". This does not necessarily provide a clear enough warning for some users.

**Describe the solution you'd like**
The warning message needs to be clearer and more informative.

**Describe alternatives you've considered**
A potentially better warning message would be:

"No individual transcripts have been identified that fully overlap the described variation in the genomic sequence. Large variants might span one or more genes and are currently only described at the genome (g.) level.”

ifokkema:
Related to:
https://github.com/openvar/variantValidator/issues/333
https://github.com/openvar/variantValidator/issues/399
https://github.com/openvar/variantValidator/issues/652

PR #752, opened 2025-09-05
Adjust warning for genomic data only variants

John-F-Wagstaff:
The original warning is only clear if you understand the limitations of the current hgvs standard, but many of our users do not! This is more wordy, but much more clear.

See https://github.com/openvar/variantValidator/issues/751 for more details


Peter-J-Freeman:
Fair enough. Yep, pull it into the code base

PR #754, opened 2025-10-06
Expanded repeat in out, moved + hgvs object input

John-F-Wagstaff:
A version of the Expanded repeat as output, when provided as input, code.  Moved from the old file names, and fixed up to use hgvs objects instead of text input (and output), to avoid costly re-parsing.

### DIFF
--- a/VariantValidator/modules/expanded_repeats.py
+++ b/VariantValidator/modules/expanded_repeats.py
@@ -87,6 +87,8 @@ class TandemRepeats:
         self.no_norm_evm = False
         self.genomic_conversion = None
         self.original_position = None
+        self.reference_sequence_bases = None
+
         # Define the wobble bases map with proper regex
         self._wobble_bases_map = {
             'A': r'A',       # Adenine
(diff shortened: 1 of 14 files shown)


Peter-J-Freeman:
@John-F-Wagstaff 
I'm getting 
FAILED tests/test_configuration.py::TestConfigValues::test_file_structure - AssertionError: Element counts were not equal:
FAILED tests/test_liftover.py::TestLiftover::test_gene_variant - AssertionError: assert SequenceVariant(ac=NC_000024.9, type=g, posedit=541732delA, rel_ac=) == 'NC_000024.9:g.541732del'
FAILED tests/test_variantformatter_inputs.py::TestVFvariantsAuto::test_issue_392b - TypeError: argument of type 'SequenceVariant' is not iterable
FAILED tests/test_variantformatter_inputs.py::TestVFvariantsAuto::test_issue_744a - AssertionError: assert {'NC_000022.1... 'ref': 'C'}}} == {'NC_000022.1... 'ref': 'C'}}}
FAILED tests/test_variantformatter_inputs.py::TestVFvariantsAuto::test_issue_744b - AssertionError: assert {'NC_000012.1... 'ref': 'C'}}} == {'NC_000012.1... 'ref': 'C'}}}


Do I need a vv_hgvs update too? Or a VF update?

John-F-Wagstaff:
Hello Pete

Sorry I mentioned it in passing when we last talked about this, but should have put it in this PR, there is a VF pull request that is needed first to make this work.

John-F-Wagstaff:
We should yes, the design should not have been changed, just the details of some of the methods, though we now can (and do) store ER variants as HGVS objects which is needed for the later VRS work.

John-F-Wagstaff:
Yes, as long as you have had a check over the code that I changed, as well as running the tests. It started off as code you wrote, so hopefully anything really bad should jump out at you, but either way it is always possible to read over errors because you know what it should do, so a second set of eyeballs over the changes is usually a good idea.

If you are fine with it then it should be OK.

Issue #767, opened 2025-12-09
Error in the `ref` base does not trigger a validation error for an intronic variant


ielis:
**Describe the bug**

Error in the reference is not picked up for an intronic variant. This happens for the following query:
- `genome_build`: `hg38`
- `variant_description`: `NM_006846.4:c.2666+5A>G`
- `select_transcripts`: `mane_select`

The 5th intronic base in fact corresponds to `G` (not `A`, as indicated in the variant description), but this is not picked up by variant validator. The gene in question (*SPINK5*) is on the forward strand.

The returned JSON document includes the following:

```
"validation_warnings": [
      "NM_006846.4:c.2666+5A>G automapped to NM_006846.4:c.2666+5="
    ]
```

However, I am not sure this is very useful.

**To Reproduce**

Execute the following query:

```shell
curl -X GET https://rest.variantvalidator.org/VariantValidator/variantvalidator/hg38/NM_006846.4%3Ac.2666%2B5A%3EG/mane_select?content-type=application%2Fjson -H 'accept: application/json'
```

**Expected behavior**

I believe, that the variant validator should complain about the error in the reference.

**Screenshots**
Not applicable

**Desktop (please complete the following information):**
Not applicable

**Smartphone (please complete the following information):**
Not applicable

**Additional context**
Not applicable

Peter-J-Freeman:
The warning should be generated based on the genomic reference sequence under consideration. However, I am unsure whether it would currently be triggered by anything but the primary assembly mapping. I will push what I have to a branch and we may need some additional examples for NT/W_(NM_)

As it stands, I think it will only warn for the NC_ auto-selected by evm for the givn genome build

Peter-J-Freeman:
I stand corrected

```python
import json
import VariantValidator
vval = VariantValidator.Validator()
variant = "NW_011332691.1(NM_012234.6):c.335+1A>G"
genome_build = 'GRCh38'
select_transcripts = 'all'
transcript_set = 'refseq'
validate = vval.validate(variant, genome_build, select_transcripts, transcript_set)
validation = validate.format_as_dict(with_meta=True)
print(json.dumps(validation, sort_keys=True, indent=4, separators=(',', ': ')))
```


```json
        "validation_warnings": [
            "NW_011332691.1:g.53828T>C: Variant reference (T) does not agree with reference sequence (C)",
[...]

PR #769, opened 2025-12-09
Changes to address issue #767

Peter-J-Freeman:


Peter-J-Freeman:
I think IntronBaseMmismatchError will be ReferenceBaseError or somethig we can use for both intronic or exonic since its the same overall issue. This sound OK?

I will add a warning as specified

> Finally it is a bit of a nit-pick, but the range based issue could probably have been submitted as a separate patch, it is functionally separate from the more complex mapping work, however this is mostly me being fussy on a style issue.

True, its me cramming too much in due to snatching come code time

Peter-J-Freeman:
@John-F-Wagstaff @ifokkema , do these address all the above comments - see latest commit

To do if agreed
- add IntronBaseMismatchError if we have seen the NC(NM) format if this would helo too?

PR #782, opened 2026-02-16
handle \s+ in VCF

Peter-J-Freeman:


John-F-Wagstaff (approved):
Looks good, a few comments but only 1 fix, which could be put off for later if needed.

Firstly: If we wanted to be more strict on inputs to-be processed you could trigger a strip on the input before testing, in vvMixinCore, but this would only eliminate broken inputs that will drop out with a error about column numbers anyway so this is fine as is.

Secondly: Testing for "\t" first then regex for '\s+' will be faster for tabs but slightly slower on the non tab containing inputs. As non-tab inputs are non-standard, and also should be less common, this looks fine to me.

Finaly: We end up not splitting on ',' in the split_vcf_line function, after this patch, given that we skip comma deliminated inputs already this also checks out, However the VcfConversionError on line 21 might need updating here though, as the text no longer matches the code. 

So you might want to get in a fix on the VcfConversionError while we are here.

Peter-J-Freeman:
@John-F-Wagstaff, this should then deal with these useful thoughts. A little additional code, but likely faster for correct \t versions

Issue #815, opened 2026-05-15
Strange SELENON p. outputs


Peter-J-Freeman:
Dear Pete/Ivo,

first, on the screenshot send by Pete I do not see a clear waning, but I
might miss it.

Reagrding strange results, try e.g. NM_020451.3:c.447dup. The predicted
consequence is p.(Asp430Ter) but VV gives me
p.(Asp150_Ser152delinsTerProTer). No clue why this is given.

For other variants, e.g. NM_020451.3:c.407del p.(Ser146Ter) or
NM_020451.3:c.532del p.(Leu179Ter) I get "p.?". This is a strange resutl
and I see no where a warning on my screen. You could say the result
"p.?" is the warning but I wonder whether users will understand this.

Best regards,

        Johan

Peter-J-Freeman:
> Reagrding strange results, try e.g. NM_020451.3:c.447dup. The predicted consequence is p.(Asp430Ter). 

This cannot be true.

Not sure this is right. Here are the translations

REF
```
MGRARPGQRGPPSPGPAAQPPAPPRRRARSLALLGALLAAAAAAAVRVCARHAEAQAAARQELALKTLGTDGLFLFSSLDTDGDMYISPEEFKPIAEKLTGSCSVTQTGVQWCSHSSLQPQLPWLNUSSCLSLLRSTPAASCEEEELPPDPSEETLTIEARFQPLLPETMTKSKDGFLGVSRLALSGLRNWTAAASPSAVFATRHFQPFLPPPGQELGEPWWIIPSELSMFTGYLSNNRFYPPPPKGKEVIIHRLLSMFHPRPFVKTRFAPQGAVACLTAISDFYYTVMFRIHAEFQLSEPPDFPFWFSPAQFTGHIILSKDATHVRDFRLFVPNHRSLNVDMEWLYGASESSNMEVDIGYIPQMELEATGPSVPSVILDEDGSMIDSHLPSGEPLQFVFEEIKWQQELSWEEAARRLEVAMYPFKKVSYLPFTEAFDRAKAENKLVHSILLWGALDDQSCUGSGRTLRETVLESSPILTLLNESFISTWSLVKELEELQNNQENSSHQKLAGLHLEKYSFPVEMMICLPNGTVVHHINANYFLDITSVKPEEIESNLFSFSSTFEDPSTATYMQFLKEGLRRGLPLLQP*
```

VAR
```
[...]

Issue #799, opened 2026-03-27
VCF SNV ouotputs


Peter-J-Freeman:
**Is your feature request related to a problem? Please describe.**
For large deletions duplications and inversions, the VCF lines become huge

**Describe the solution you'd like**
convert them into

- chr-start-end-del
- chr-start-end-inv
- chr-start-end-dup (VCF spedification does not as far as I am aware define between tandem and unknown location dup)

**Describe alternatives you've considered**

**Additional context**

Peter-J-Freeman:
Currently set to range = 50
@ifokkema @John-F-Wagstaff @leicray comments on this please.

The output look like 

```python
import json
import VariantValidator
vval = VariantValidator.Validator()
variant = "chr1\t1000000\t.\tN\t<DEL>\t.\tPASS\tSVTYPE=DEL;END=1005000"
genome_build = 'GRCh38'
select_transcripts = 'all'
transcript_set = 'refseq'
validate = vval.validate(variant, genome_build, select_transcripts, transcript_set, lovd_syntax_check=True)
validation = validate.format_as_dict(with_meta=True)
print(json.dumps(validation, sort_keys=True, indent=4, separators=(',', ': ')))
```

```json
"hg38": {
                "hgvs_genomic_description": "NC_000001.11:g.1000000_1005000del",
                "vcf": {
                    "alt": "DEL",
                    "chr": "chr1",
                    "pos": "1000000",
                    "ref": "1005000"
                }
            }
```

I am aware that ref is now an integer, but I do not want to change the structure of the dicts

My question is, what should the threshold be for converting to this format. Shorter works for us, but I am aware of dbSNP chr-pos-ref-alt formats of >500 bases. But a reason for us to do this is display and also database capacity

Peter-J-Freeman:
before application

<img width="1396" height="638" alt="Image" src="https://github.com/user-attachments/assets/605af397-f07a-476c-a48b-18eabd737d3d" />

After application

<img width="879" height="604" alt="Image" src="https://github.com/user-attachments/assets/b3801e52-0c2f-402b-b47d-b9e96ecec5c8" />

In terms of display and data storage, no brainer :)

leicray:
Looks good to me.

ifokkema:
Sorry, I completely overlooked the email about this.

[...]

PR #817, opened 2026-05-18
Develop

Peter-J-Freeman:


Peter-J-Freeman:
@John-F-Wagstaff recommend merging this then bringing in the VV 4.0.0 coede changes

Peter-J-Freeman:
@John-F-Wagstaff Can you give this a quick sanity check please. Seems to be all OK. All tests passing and additional tests included to cover a wide range of Selenon Protein Variants

VariantValidator/modules/vvMixinInit.py
```diff
@@ -678,178 +783,183 @@ def _remake_unc(prot,nucleotide_not_equal=False):
...
         # Complete variant description
-
-        # Write the HGVS position and edit
-        # start by handling delins->ins transitions from the cds to prot mapping
-        if not pro_inv_info['prot_del_seq']:
-            # must be != exclusive coordinates
-            assert pro_inv_info['edit_start'] != pro_inv_info['edit_end']
-            from_aa = prot_ref_seq[pro_inv_info['edit_start']]
-            to_aa = prot_ref_seq[pro_inv_info['edit_end']]
+        if posedit:
```
John-F-Wagstaff: For future code, you could have returned early here, doing so would have avoided all the later indenting that inflated the change log.  

It is also best to do so whenever you can as it simplifies the later logic, and makes things more readable. Not only does it make it clear when you have finished but python's use of indenting to handle code blocks also becomes less and less readable the more indented code becomes. 

Fixing this now would probably involve re-building the pull request, which I can do if you want, but is not necessary if you don't want to.  I am mostly mentioning this as you seem very reluctant in your other code to return early at all, when you probably want to return early as a normal way of handling things (e.g. cases like this, simple error/null input cases ect.).

John-F-Wagstaff (approved):
OK, this looks good, the only comment I have is more on style, though it would have also made checking which bits of the logic changed easier, so I think this is fine if you want to pull it.

VariantValidator/modules/vvMixinInit.py
Peter-J-Freeman: Happy for you to refactor it, and yes, I will take the lesson and look to return earlier. Thanks for the nidge. Are you OK to do it? Might as well keep it clean where possible.

John-F-Wagstaff: I should be OK to do it, I will have a try.

Peter-J-Freeman: Let me know if you get stuck and I can assist too. Thanks

Peter-J-Freeman:
@John-F-Wagstaff. While you are refactoring, can you see if there is a way of passing a warning to the users that the variant is in a Seleno Protein encoding gene, so the codon `UGA` is being translated to `Sec` and not `Ter`. That would get around comments from Johan that there are no warnings. I didnt think to warn for successful validations, but perhaps for educational purposes, this might be useful to our users

VariantValidator/modules/vvMixinInit.py
John-F-Wagstaff: It looks OK to do, but I will have to do a force push on develop to do this.  It would replace all commits from the merge commit for the #815 branch onwards, with commits based on a fixed-up version of the merge commit. This is doable but is is often seen as rude as it can impact other users/branches. 

Given how few other people directly get our code from git, let alone from develop this might be OK still, but it is only sensible if you don't have anything based on commits after the merge.  The tests seem OK so far, but is this alright by you? (and if so do you have anything based on dev post the #815 merge.)

Peter-J-Freeman: Nope. It can go. Thanks

Did you see the message about adding a warning in the pull request main area? Might as well add that too if it's not too tricky.

John-F-Wagstaff: It should be fine, but the fixup to the older commit is more "risky" so I am getting that done and tested first, then doing the extra as a commit on the end to add warnings, as followup.

Would you prefer ProteinTranslationError: or ProteinTranslationWarning: or ProteinTranslationInfo: as a prefix?

Peter-J-Freeman: I think Info technically. It's not really a warning and we tend to use warning for errors which we fix. 

Thanks

Peter-J-Freeman: @John-F-Wagstaff let me know when this is good to merge. I will do nothing until you say go :).

John-F-Wagstaff: If you want to merge into dev rather than wait go ahead, I ran into some problems with tests not getting started with test_, and so not being run by pytest, that need fixing before I complete this. Unless you are messing with vvMixinInit it should only slightly increase the time to finish the fixup. Any branches based on commits post 
commit: dc0bfa9b3ee50c0eb831bf05d5bb445edd53be59 (fix uncertain position parsing of extracted variants errors)
are probably best merged and cherry-picked etc by me while I do the rest. Especially as you will want to not have anything active based on such commits when or after I do the force push.

Peter-J-Freeman: no rush. I would rather wait until refactored. I have nothing to push to dev right now. Will you be making the tweaks we identified here as part of a larger push to develop (which includes moving the VCF shorthand code movement of location and other stuff we discussed) or in parts, i.e. the refactoring of this then the additional parts. Just planning when to address the remaining parts of https://github.com/openvar/variantValidator/issues/799#issuecomment-4563615372 and making sure I dont update some thing that will clash. Would rather wait for you to say go and what to do next i.e. when I can apply the changes. there is no rush at all.

John-F-Wagstaff: OK.

I was just planning on getting the test issues fixed as they impact the Selenon code, then pushing this fix as is, the less I mess with things by re-basing on an non-private branch the better. I may even skip some of the test fixes until after.

Peter-J-Freeman: Sounds good, so push the selenon stuff in this commit with the test fixtures working. Do you want to include the edits to the VV VCF shrinking output in your work then, and I will plug it into the web interface? The model is shown in the link. We also want to move the point where VCF shrinking happens from 50bp to 500. 

Model for VV dict is 

```
    "vcf": {
        "alt": "DEL",
        "chr": "chr1",
        "pos": "1000000-1005000",
        "ref": "N"
    }
```

I think the pos in the dict is already a string, but we will have to enforce this behaviour.

John-F-Wagstaff: I have pushed all the changes needed to get this back to where we want it. I will change my current branch of the mapping fixes to match the new VCF format and make pull requests into dev ASAP.

When you try to pull from this git will recommend merging or re-basing, *don't*! 

If you have no local changes the you just need to delete the last few commits from your local branches to get this to pull correctly. The other options will make a lot of mess. Given this it is probably best to delete and pull the new version on all affected copies of develop, otherwise if you use an old version as the basis for new work it will cause issues to untangle it later. If you do have local commits (or branches) based on this contact me and we can work the best method out, though for commits it will probably just want git branch followed handling develop as above, followed by git cherry-pick from the backup branch you just created with the extra changes.

Peter-J-Freeman: Done, I deleted my local branch and re-pulled from develop.

PR #824, opened 2026-06-04
Develop

Peter-J-Freeman:


VariantValidator/modules/vvMixinInit.py
```diff
@@ -417,22 +417,47 @@ def _remake_unc(prot,nucleotide_not_equal=False):
...
-                or re.search(r'\d+\+', str(hgvs_transcript.posedit.pos)) \
-                or re.search(r'\*', str(hgvs_transcript.posedit.pos)) \
-                or (re.search(r'[cn].-', str(hgvs_transcript)
-                              ) and "dup" not in hgvs_transcript.posedit.edit.type) or (
-                ("dup" in hgvs_transcript.posedit.edit.type and
+        logger.info("Test for intronic and UTR trapping")
+        if (
+                # Intronic variants (HGVS: position+offset or position-offset)
+                # e.g. 123+2, 456-1 → always trap
+                re.search(r'\d+-\d+', str(hgvs_transcript.posedit.pos))
```
John-F-Wagstaff: Nit Pick:
I left the old regular expression based pattern in all over our code, when switching over from text to hgvs objects, in order to avoid fattening the diff and adding to the existing complexity of that switch.  

Now that this switch is completed though we probably want to swap out for the more efficient methods that don't rely on converting to a string then running a regular expression, whenever we do edits.

For this we should be able to detect intronic variants via 
`hgvs_transcript.posedit.pos.start.offset or hgvs_transcript.posedit.pos.end.offset 
`
As any offset state other than 0 means intronic

edit: if we have simple positions mixed in then `getatr(hgvs_transcript.posedit.pos.start, "offset", False` (and the same for the end) would work instead 

(this probably is not worth changing here but it may be worth noting for future code)

Peter-J-Freeman: Totally agreed. I tried to do the same here to minimise the regex need. Let me see if I can get it tighter

VariantValidator/modules/vvMixinInit.py
```diff
@@ -417,22 +417,47 @@ def _remake_unc(prot,nucleotide_not_equal=False):
...
+                or re.search(r'\d+\+\d+', str(hgvs_transcript.posedit.pos))
+
+                # Structural variants (dup/del/inv/ins/delins)
+                # Allow through if they extend outside clean CDS boundaries:
+                # - end contains "-" → extends toward 5′ direction
+                # - start contains "*" → extends into 3′ direction
+                or (
+                (any(x in hgvs_transcript.posedit.edit.type for x in ["dup", "del", "inv", "ins"]) and
```
John-F-Wagstaff: Low priority nitpick:

This line relies on `x in `to avoid having to specify delins, maybe also as a holdover from this being a string hgvs in older versions, but as a cost does an sub-sting in string comparison per test, 
switching from:
`(any(x in hgvs_transcript.posedit.edit.type for x in ["dup", "del", "inv", "ins"])
`
to 
`hgvs_transcript.posedit.edit.type in ["dup", "del", "inv", "ins", "delins"]
`
should be a bit quicker and is more straightforward  if you have code like 
`"dup" in hgvs_transcript.posedit.edit.type
`
you probably want 
`hgvs_transcript.posedit.edit.type = "dup" 
`
as well, to avoid the string search

VariantValidator/modules/vvMixinInit.py
```diff
@@ -417,22 +417,47 @@ def _remake_unc(prot,nucleotide_not_equal=False):
...
                 or
-                ("dup" in hgvs_transcript.posedit.edit.type and
+                (any(x in hgvs_transcript.posedit.edit.type for x in ["dup", "del", "inv", "ins"]) and
                  "*" in str(hgvs_transcript.posedit.pos.start))
-                ):
+            )
+
+                # Fully 3′ UTR variants (DOWNSTREAM)
+                # Trap ONLY if BOTH start and end are "*" positions
+                or (
```
John-F-Wagstaff, line 450: Low priority nitpick:
For the UTR logic we can do 2 tests, to avoid string conversion + regex testing:

firstly testing for 3' utr uses
`hgvs_transcript.posedit.pos.start.base < 0 `

Secondly however testing for the 5' utr is a bit more fancy, we have to add
`from vvhgvs.enums import Datum`
to our imports, then test for
`Datum.CDS_END = hgvs_transcript.posedit.pos.start.datum`

I think that all of our transcript mappings should be the more complex position type that always have this, but if not 
testing for `type = 'c'` (which definitely always will) or using `getattr()` may be needed to avoid complex try catch logic.

VariantValidator/modules/vvMixinInit.py
```diff
@@ -618,6 +649,63 @@ def _remake_unc(prot,nucleotide_not_equal=False):
         logger.info(f"VarSeq: {prot_var_seq}")
         logger.info(f"pro_inv_info: {pro_inv_info}")
 
+        ######################################
+        #  Ter codon interruption frame shifts
+        ######################################
+
+        if (
+                ("del" in str(hgvs_naughty.posedit.edit)
```
John-F-Wagstaff, line 664: The logic looks good here, but the same low priority nitpicks apply to this if statement (i.e. `hgvs_naughty.posedit.edit.type in ['del', 'delins']` instead of `"del" in` and `==` not `in` for `'ins'`)

VariantValidator/modules/vvMixinInit.py
```diff
@@ -396,23 +396,34 @@ def _remake_unc(prot,nucleotide_not_equal=False):
...
             try:
                 shifts = evm.c_to_p(hgvs_transcript)
                 shifts = _remake_unc(shifts,nucleotide_not_equal=nucleotide_not_equal)
-                if "identity" in shifts.posedit.edit.type:
-                    not_delins = False
-                if 'del' in shifts.posedit.edit.type or 'dup' in shifts.posedit.edit.type:
-                    not_delins = False
                 if "fs" in shifts.posedit.edit.type:
                     not_delins = True
+                if "ext" in shifts.posedit.edit.type:
```
John-F-Wagstaff: Low priority nitpick as before: For this and the else block below, we probably want to shift to avoid `in` on a string when we can use `==` or `in` on a list of strings (which is equivalent to testing `==` on each member of the list)

tests/test_inputs.py
John-F-Wagstaff, line 1: Some people prefer keeping a clean history for tests, if we want to follow this then merging this commit with the previous commit would be the way to go, I don't mind either way but it can help with some automatic tooling for git bisecting.

VariantValidator/modules/vvMixinCore.py
```diff
@@ -1035,6 +1035,21 @@ def validate(self,
                                                 r'\(LRG_.+?\)', '', predicted_protein_variant.ac)
                                 else:
                                     format_lrg = None
+
+                                # convert UTR variants from p.? to p.(?)
+                                try:
+                                    if (re.match("^-\d+",str(variant.hgvs_coding.posedit.pos.end)) or
```
John-F-Wagstaff: Low priority nitpicik: this could also be tweaked to avoid the regex if we wanted to.

John-F-Wagstaff (approved):
There is not anything high priority here, so it should be fine to go in as-is, otherwise @Peter-J-Freeman If you want me to have a quick check over after any refinements then that would be fine.

Peter-J-Freeman:
Working on the refinements so I know how to use them. Will push up the fixes for a review and any I missed you can polish off

John-F-Wagstaff:
Sounds good, we need to go over the whole code base at some point, but it is such a big job, and should not change the logic, so we haven't had a good enough reason to prioritise it.

Peter-J-Freeman:
@John-F-Wagstaff should be good to go now. final quick check pls

John-F-Wagstaff:
Looks good, as far as I can see.

PR #825, opened 2026-06-04
feat: update UTR searching to use datum instead of regex

Peter-J-Freeman:
replaces string-based checks with HGVS-native attribute logic. This is a prototype and needs to be rolled out widely

### DIFF
--- a/VariantValidator/modules/vvMixinCore.py
+++ b/VariantValidator/modules/vvMixinCore.py
@@ -1,6 +1,7 @@
 import vvhgvs
 import vvhgvs.exceptions
 import vvhgvs.normalizer
+from vvhgvs.enums import Datum
 from vvhgvs.location import Interval
 from vvhgvs.sequencevariant import SequenceVariant
 from vvhgvs.posedit import PosEdit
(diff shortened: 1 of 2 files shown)


VariantValidator/modules/vvMixinCore.py
```diff
@@ -1038,14 +1039,23 @@ def validate(self,
 
                                 # convert UTR variants from p.? to p.(?)
                                 try:
-                                    if (re.match("^-\d+",str(variant.hgvs_coding.posedit.pos.end)) or
-                                            "*" in str(variant.hgvs_coding.posedit.pos.start)):
-                                        logger.info(f"CDS variant {variant.hgvs_coding} identified. "
-                                                    f"Updating from p.? to p.(=)")
+                                    is_5utr = (variant.hgvs_coding.posedit.pos.end.datum ==
```
John-F-Wagstaff: You could shorten this even more, if you want. 
Since all non `CDS_START`, including both `CDS_END` and the n type `SEQ_START` must be +ve you can just test for -ve
i.e 
`is_5utr = variant.hgvs_coding.posedit.pos.end.base < 0
`
alternatively
```
is_utr = (variant.hgvs_coding.posedit.pos.end.base < 0 or variant.hgvs_coding.posedit.pos.start.datum == Datum.CDS_END and )
```
or just the stuff in the brackets instead of both ` is_5utr or is_3utr`

VariantValidator/modules/vvMixinInit.py
```diff
@@ -435,36 +429,33 @@ def _remake_unc(prot,nucleotide_not_equal=False):
...
-                # - end contains "-" → extends toward 5′ direction
-                # - start contains "*" → extends into 3′ direction
+                # Intronic variants: offset != 0  (formerly + or - in string)
+                (hgvs_transcript.posedit.pos.start.offset != 0 or
+                 hgvs_transcript.posedit.pos.end.offset != 0)
+
+                # Structural variants extending outside CDS:
+                or (
+                hgvs_transcript.posedit.edit.type in ["dup", "del", "inv", "ins", "delins"] and
+                hgvs_transcript.posedit.pos.end.datum == Datum.CDS_START and
```
John-F-Wagstaff, line 441: Again you could trim the Datum.CDS_START here as only  Datum.CDS_START can be -ve

VariantValidator/modules/vvMixinInit.py
```diff
@@ -435,36 +429,33 @@ def _remake_unc(prot,nucleotide_not_equal=False):
...
             )
 
-                # Fully 5′ UTR variants (UPSTREAM)
-                # Only trap if BOTH positions START with "-"
-                # Avoids intronic cases like 123-4
+                # Fully 5′ UTR (both negative relative to CDS_START)
                 or (
-                str(hgvs_transcript.posedit.pos.start).startswith('-')
-                and str(hgvs_transcript.posedit.pos.end).startswith('-')
+                hgvs_transcript.posedit.pos.start.datum == Datum.CDS_START and
```
John-F-Wagstaff: Another possible trimming available here as above

John-F-Wagstaff (approved):
Looks good so far, you could go for it with this, but there are a few more easy gains, if you want to put in another commit, or use `git commit --amend ` and force push.

PR #828, opened 2026-06-08
Known bad alignment

Peter-J-Freeman:


VariantValidator/modules/gene2transcripts.py
```diff
@@ -534,6 +543,61 @@ def gene2transcripts(g2t, query, validator=False, bypass_web_searches=False, sel
...
+    not tx_for_region which has a different list structure.
+
+    Examples:
+        NM_000088.4 -> ('NM_000088', 4)
+        ENST00000225964.10 -> ('ENST00000225964', 10)
+        ENST00000486572.1/GRCh38 -> ('ENST00000486572', 1)
+    """
+
+    # remove optional genome build suffix
+    accession = accession.split("/")[0]
```
John-F-Wagstaff: If this is the genome build suffix that turns up in VVTA on some Ensembl data it is not optional, but instead a sign that the same seq ID with the same version number was published for both GRCh37 and GRCh38 and is different between them despite the identical version number. 

If this is the cause then you can probably do a simplified version for RefSeq. You might want to do more if you are making that change (or some other tweak given this issue), as although it is not exactly worth changing unless you make more changes but partition/rpartition is  quicker than split. I have no idea why it is apparently even faster than split limited to 1 split point, (posibly as a consequence of being less flexible?) but it works whenever you want the first (or last for rpartition) match to be the split point.

John-F-Wagstaff:
Looks OK, just a question on the Suffixes, and a possible simplification. I would just hit approve but I think that the suffix issue needs thinking on at least first.

I do wonder about doing a latest transcripts only materialised view for VVTA, but how to deal with the differing (suffixed) Ensembl data is a fun question, and we would then need to plumb it in correctly later in multiple places.

Peter-J-Freeman:
will fix this up, but I have a feeling that there are other instances in VV and VF where we need to fix too

John-F-Wagstaff:
I believe that the intention was to just not use them for normal output, mostly, but to keep them to have them to map from if user input matched, with appropriate warnings, if we wanted to handle this later. But we never did get to being able to handle that. Not only as the ID for the user input would not match the ID for the ensemble data, so in that case so we would have to check specifically for this issue, but also because we would then have to handle the possible inputs interpretations for GRCh37/38 and make sure they pass through the rest of the code properly.

As such we probably just need to make sure nothing is crashing, for now, since this would be bonus feature development, not bug-fixing.

I think the intention was that if we got around to this to find both suffixed and resubmit the 2 suffixed with warnings?, e.g `"TranscriptVersionWarning: Two versions of this ENSEMBL transcript exist with different sequence but the same ID and version number, one for GRCh38 and one for GRCh37. Neiter of these should be used in publications, but both versions have been resubmitted. Please check the output for possible discrepancies and select an appropriate new variant definition, you may need to use a RefSeq transcript instead."` but only for inputs that used an ENS id that matched the transcript, not for genomic or gene id based queries)

Peter-J-Freeman:
hope this push maintains the required format. We will add the other stuff to the ever expandin to do list :)

VariantValidator/modules/gene2transcripts.py
```diff
@@ -546,24 +546,28 @@ def gene2transcripts(g2t, query, validator=False, bypass_web_searches=False, sel
...
+        ENST00000225964.10
+            -> ('ENST00000225964', 10, None)
+
+        ENST00000486572.1/GRCh38
+            -> ('ENST00000486572', 1, 'GRCh38')
     """
 
-    # remove optional genome build suffix
-    accession = accession.split("/")[0]
+    accession_part, _, genome_build = accession.partition("/")
```
John-F-Wagstaff: This looks OK, if we do want to use this, but as it it just the broken ensembl data we might want to just dump the whole line for these depending, we treat them as effectively masked for bad quality elsewhere anyway.

VariantValidator/modules/gene2transcripts.py
```diff
@@ -574,25 +578,29 @@ def filter_latest_transcripts(rows):
     # remove blat rows
     rows = [row for row in rows if row[5] != "blat"]
```
John-F-Wagstaff: if we want to filter out the bad data we could just do a
` rows = [row for row in rows if '/' not in row[3] ]`
or if you pass transcript set along (which would require a few more changes)
```
if transcript_set == "ensembl":
     rows = [row for row in rows if '/' not in row[3] ]

```
Then we would be down to only 1 split and 1 less comparison per filter, though again this only works if we want to treat these as "bad quality don't use" data as opposed to "bad but still usable".

Peter-J-Freeman: I spotted an additinal loop that cold be removed as well, so these edits remove a loop and also remove other known issues. We now have a filter that other hdp.get_tx_for_gene can use to clean up. Can likely be adapted for get tx_by region or overlapping variant etc.

John-F-Wagstaff (approved):
Looks good now if you want to push it.

Issue #810, opened 2026-04-29
Some impossible mappings


ifokkema:
**Describe the bug**
I'm trying to wrap my head around some impossible mappings, and I am unable to understand WTF is going on... In short, I'm trying to solve the problem where one given variant returns multiple mappings on the other build, caused by multiple transcripts mapping elsewhere. While trying to resolve this using what I thought was logical, I got stuck completely.

**To Reproduce**
[This GRCh37-based duplication](https://rest.variantvalidator.org/LOVD/lovd/GRCh37/NC_000009.11%3Ag.95237063_95237068dup/refseq/mane_select/tx/primary?content-type=application%2Fjson) returns two GRCh38 mappings. Summarized output:
```json
{
  "NC_000009.11:g.95237063_95237068dup": {
    "NC_000009.11:g.95237063_95237068dup": {
      "g_hgvs": "NC_000009.11:g.95237063_95237068dup",
      "hgvs_t_and_p": {
        "NM_001012267.3": {
          "gap_statement": null,
          "primary_assembly_loci": {
            "grch37": {
              "NC_000009.11": {
                "hgvs_genomic_description": "NC_000009.11:g.95237063_95237068dup"
              }
            },
            "grch38": {
              "NC_000009.12": {
                "hgvs_genomic_description": "NC_000009.12:g.92474781_92474786dup"
              }
            }
          },
          "select_status": {
            "mane_select": true
          },
          "t_hgvs": "NM_001012267.3:c.564+94922_564+94927dup"
        },
        "NM_017680.6": {
          "gap_statement": "NM_017680.6 contains 3 fewer bases between c.152_153 than NC_000009.11",
          "primary_assembly_loci": {
            "grch37": {
              "NC_000009.11": {
                "hgvs_genomic_description": "NC_000009.11:g.95237063_95237068dup"
              }
            },
            "grch38": {
              "NC_000009.12": {
                "hgvs_genomic_description": "NC_000009.12:g.92474784_92474786del"
              }
            }
          },
          "select_status": {
            "mane_select": true
          },
          "t_hgvs": "NM_017680.6:c.150_153-1dup"
        }
      }
    }
  }
}
```

[...]

PR #836, opened 2026-06-17
bug: prelim checks for complete repeat block dups

Peter-J-Freeman:
Expanded repeat blocks where the gap is exactly 1
repeat block with respect to the transcript cause
vv_hgvs and VV to miss duplicaations at the g>t
conversion. Refer to #810. The patterns can be
identified as shown in the code. However,
the conditional if statements may need to be
expanded with additional variants over time.
We also need to identify similar variants in
antisense genes wrt to the chr reference. However
this sets the framework and fixes the test variant in the linked issue


VariantValidator/modules/gapped_mapping.py
```diff
@@ -2,13 +2,12 @@
 import re
 import logging
 import vvhgvs.exceptions
-from . import utils as fn
+from sqlparse.utils import offset
```
John-F-Wagstaff: We don't want this imported

John-F-Wagstaff:
My brain is a bit fried today to go into the details on this one.  I need to work through it step-wise to check the logic on the more complex conditional(pair), which could take a while.

Peter-J-Freeman:
Yeah, these are a pain.

FYI, the reason these are not merged in the hard push code, the variants overlap, so cannot be merged by the current tool, so it goes through the path you see where it is spotted.

Peter-J-Freeman:
@John-F-Wagstaff The weird import was not used. Removed. And I kept import utils as fn removed as well as it did not turn up in the file. Only hgvs_utils was used. I think it was a PyCharm co-pilot autofill error. I think I hit tab and not delete when I originally removed the import!

The tests show that the same dup location, shorter and longer round trip correctly

Peter-J-Freeman:
I will see if I can find an antisense example next!

Peter-J-Freeman:
@John-F-Wagstaff antisense checks now added too

John-F-Wagstaff (approved):
Looks a lot better with a few tests, sorry that I did not have time to do a deep dive on the logic, but it should be good to go in now if you want.

Peter-J-Freeman:
no worries, I now need to figure out how to resolve the conflicts. Cant do it on the git editor

PR #837, opened 2026-06-17
bug: Make batch_list local not global

Peter-J-Freeman:
Thought. Is batch list added to to the Validator object causing race conditions. Difficult to test, but we can move batch_lisy to local scope anyway.


Peter-J-Freeman:
@John-F-Wagstaff. Despite the bad commit message, is this one good to go?

John-F-Wagstaff (approved):
This looks good to go in when you want to.

PR #838, opened 2026-06-17
Global norm and map

Peter-J-Freeman:
PR merges in the use of global normalizers and mappers

@John-F-Wagstaff, tests for exon boundary crossing in expanded_repeats.py
are failing. It is not clear to me how to pass the global mappers correctly.

### DIFF
--- a/VariantValidator/modules/vvMixinCore.py
+++ b/VariantValidator/modules/vvMixinCore.py
@@ -11,7 +11,6 @@ import sys
 import logging
 import json
 import time
-from vvhgvs.assemblymapper import AssemblyMapper
 from VariantValidator.modules import hgvs_utils
 from VariantValidator.modules import utils as fn
 from VariantValidator.modules import vvMixinConverters
@@ -187,29 +186,9 @@ class Mixin(vvMixinConverters.Mixin):
                         toskip = self._get_transcript_info(my_variant)
                         if toskip:
                             continue
-                        # Create easy variant mapper (over variant mapper) and splign locked evm
-                        my_variant.evm = AssemblyMapper(self.hdp,
-                                                        assembly_name=primary_assembly,
-                                                        alt_aln_method=self.alt_aln_method,
-                                                        normalize=True,
-                                                        replace_reference=True
-                                                        )
 
-                        # Setup a reverse normalize instance and non-normalize evm
-                        my_variant.no_norm_evm = AssemblyMapper(self.hdp,
-                                                                assembly_name=primary_assembly,
-                                                                alt_aln_method=self.alt_aln_method,
-                                                                normalize=False,
-                                                                replace_reference=True
-                                                                )
-
-                        # Create a specific minimal evm with no normalizer and no replace_reference
-                        my_variant.min_evm = AssemblyMapper(self.hdp,
-                                                            assembly_name=primary_assembly,
-                                                            alt_aln_method=self.alt_aln_method,
-                                                            normalize=False,
-                                                            replace_reference=False
-                                                            )
+                        # Configure mappers
+                        my_variant.get_mappers(self)
 
                     if my_variant.reftype in [':c.', ':n.']:
                         my_variant.gene_symbol = self.db.get_gene_symbol_from_transcript_id(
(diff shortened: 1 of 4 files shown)


Peter-J-Freeman:
@John-F-Wagstaff , see above for failing tests in expanded_repeats.py, othewise this is passing all tests and knocked about 23 seconds off testing

John-F-Wagstaff (commented):
Looks good so far, but needs the final fixes for expanded repeats
@Peter-J-Freeman  do you want me to look at getting the expanded repeats plugged in myself and commit onto this branch? or just dig up the issue and pass you the howto?

Peter-J-Freeman:
If you could plug it in while making your edits to global variables please. It looks to me like it might be related to where things end up being placed, so maybe better to pull this branch and continue the restructure on it.

VariantValidator/modules/vvMixinInit.py
```diff
@@ -188,42 +196,86 @@ def __init__(self):
...
+        # VV selected tools
+        self.reverse_hn = None
+        self.hn = None
+        self.merge_normalizer = None
+        self.reverse_merge_normalizer = None
+
+
+
+    # Create additional normalizers
+    def create_additional_normalizers_and_mappers(self):
```
John-F-Wagstaff: So I skipped past this code on my first read through, but having had more time to think we need to replace this.

We can't have a self.reverse_hn, self.hn, self.merge_normalizer, or self.reverse_merge_normalizer set globally on the Validator object, from a self.alt_aln_method because self.alt_aln_method should not exist as a global Validator object variable.  We either need to set them on the Variant object (as pointers to the already created versions exactly as we do here) or use them based on the state of self.alt_aln_method at the usage points (leaving me to clean that up later, or keep them as local variables for the main validation function in vvMixinCore and pass them around that way.

I could do this as part of settings de-globalisation followup to this though if we need to.

John-F-Wagstaff: If you want to leave in the Variant object specific ones above ( in VariantValidator/modules/variant.py )‎ then just removing all of these in favour of those might be the best method.

VariantValidator/modules/vvMixinInit.py
```diff
@@ -240,10 +331,8 @@ def __init__(self):
         self.merge_normalizer = None
         self.reverse_merge_normalizer = None
 
-
-
     # Create additional normalizers
-    def create_additional_normalizers_and_mappers(self):
+    def get_normalizers(self):
```
John-F-Wagstaff, line 335: These probably all need changing as with the Mappers, they either need to be added to the Variant, made local to the main validation function or removed in favor or checking  self.alt_aln_method at the usage point (for me to fix the  self.alt_aln_method usage later)

John-F-Wagstaff:
> If you could plug it in while making your edits to global variables please. It looks to me like it might be related to where things end up being placed, so maybe better to pull this branch and continue the restructure on it.

Depending on how you want to resolve the other comments you could just put in a call to set the mappers on the Variant before calling this code, and then use the mappers from the Variant instead of creating them in this file.

Peter-J-Freeman:
@John-F-Wagstaff. Commit messages fixed up

This is where I was confused. The validator objects are set up early in the variant object. I could not figure out how
to pass them into the expanded repeat object. Can you please take a look and see what I'm missing?

VariantValidator/modules/vvMixinInit.py
Peter-J-Freeman: Makes sense, so move them into the variant object. I did wonder. This is why I started to think that it makes sense to add it to the global vs local code restructuring. 

The framework is in place at least. Yes, I think it makes sense now for you to take over and move things in terms of global, vs local, vs validator object vs variant object.

VariantValidator/modules/vvMixinInit.py
Peter-J-Freeman: Yes, I think again it makes sense to let you take over now. The framework is in place. Its all about restructuring now

VariantValidator/modules/vvMixinCore.py
John-F-Wagstaff, line 157: If you want the mappers in place during the original parsing of the input (which is needed for expanded repeats) then they need to be set here. This is also where the set appropriate normaliser code for the Variant object needs to be triggered (instead of the new normaliser creation that we have below this line).

VariantValidator/modules/vvMixinInit.py
John-F-Wagstaff: If you have decided that you want to leave this code here then that is fine, but if it is OK I would rather add a final patch to this branch, to get it all working as is, and pull that, rather than merge this in to develop with broken expanded repeats.

PR #846, opened 2026-06-25
Improve warnings is transcript selection errors

Peter-J-Freeman:
This commit adds error codes to existing transcript selection warnings. It also now warns if a transcript variant is submitted with a non-matching transcript selection, rather than the old behavior which resulted in a lost result and could become empty.


John-F-Wagstaff (approved):
That looks good, if Raymond is happy with the output then this is good to go in.

Peter-J-Freeman:
I took Raymond's guidance. So will pull it in

PR #854, opened 2026-07-01
Patch db update

Peter-J-Freeman:

PR #855, opened 2026-07-03
Vrs output

John-F-Wagstaff:
Initial version of the VRS output code for VV, valid for VRS 2.1.

@Peter-J-Freeman Please look through the output structure with an eye to whether we want to change it before it goes in, as well as checking over the code.

### DIFF
--- a/VariantValidator/modules/complex_descriptions.py
+++ b/VariantValidator/modules/complex_descriptions.py
@@ -2,8 +2,10 @@ import re
 import copy
 from vvhgvs.assemblymapper import AssemblyMapper
 from VariantValidator.modules import utils as vv_utils, format_converters
-from VariantValidator.modules.hgvs_utils import hgvs_obj_from_existing_edit
+from VariantValidator.modules.hgvs_utils import hgvs_obj_from_existing_edit,\
+        _hgvs_offset_pos_from_str_in
 import vvhgvs.exceptions
+from vvhgvs.enums import Datum, ValidationLevel
 from vvhgvs.location import Interval
 
 class UncertainConversionError(Exception):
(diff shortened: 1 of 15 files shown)


John-F-Wagstaff:
OK so the test failure is for ccds_ids

I got ` ['CCDS14231',  'CCDS14232',  'CCDS14234',  'CCDS14230',  'CCDS55394',  'CCDS94585',  'CCDS14233', 'CCDS48091', 'CCDS94586', 'CCDS14229',  'CCDS55395']` the test machine gets `[]` for variant input of 'NC_000023.10:g.33229673A>T' via 'NM_000109.3:c.7+127703T>A' should I just purge the check or is this something we want to fix?

Peter-J-Freeman:
Hi John,

I will look properly on Monday. Bit busy today

for the failing test, it is not a critical assert, so perhaps just amend the test so it passes???

John-F-Wagstaff:
OK, I will get it fixed and pushed up before then.

It is not a super hurry, I just finally wanted to get it completed and submitted. This is high priority, but more urgent supposedly shorter jobs have kept on turning up, often right as I was getting into the swing of things, and breaking my metaphorical stride. As a result this code has been hanging over me as a "should complete ASAP" for far too long.  It is a relief to finally sign off on it as done.

Peter-J-Freeman:
Dont go to a huge amount of effort. Just remove the offending assert if possible and commit and push. No need to do full pulling down of the commit etc. Or just remove the failing test unless it adds very needed fucntionality

Peter-J-Freeman:
We are gonna have to plan a merge strategy too. As I want to pull this in. 

https://github.com/openvar/variantValidator/pull/854

Lots of bugs identified and squashed

John-F-Wagstaff:
For merging strategy, I suggest that we merge  #854 first then you tell me. I will re-base this code on top of that. Only after that will we finalise this.

You have added a lot of tests to the main VV code base so any issues caused by the first two patches in this set are best examined by moving this "after" that set in git history terms. 

edit: It also might settle some of the intermittent DB connection issues, which just affected the tests.

PR #861, opened 2026-07-14
Mkdocs

Peter-J-Freeman:


Peter-J-Freeman:
@John-F-Wagstaff This pull adds

- Documentation. This will build when pulled into Master and the docs will appear at https://openvar.github.io/variantValidator/
- Error codes. I have created an error code mapper, so pretty much all errors and warnings now get coded. See docs/user-manual/features and the md is in there
- CLIs for VV, VF and gene2transcripts
- Small changes to VV and VF for consistency of variable naming with some early dropout raises. 

to do.

Create a LOVS HGVS Syntax checker Python CLI wrapper. Add the docs for the tools and then we can get @leicray to review the docs.

VariantValidator/modules/utils.py
```diff
@@ -841,6 +841,641 @@ class ObsoleteSeqError(Exception):
...
+
+    for warning in warnings:
+        warning = str(warning)
+
+        # Already coded
+        if _ALREADY_CODED.match(warning):
+            output.append(warning)
+            continue
+
+        for search, (code, replacement) in _WARNING_LOOKUPS:
```
John-F-Wagstaff: This does work, but:

Performance wise we might want to split this into searches that work with `warning.startswith(search) ` (which is a lot faster if it is valid) and those that need a full `search in warning`, unfortunately performance here may also be affected by whether the most frequent warnings turn up first in the list. 

We probably also want to go around and do a whack-a-mole on the internally generated errors.  Every location where we create our own error that we remove reduces the overhead, and if we can remove a class of warning/info/error message form the search list then every warning searched will be on average slightly faster (though again this depends on order as above). Possibly this is worth adding as a Feature request/ code cleanup type issue? (i.e. for future work)

John-F-Wagstaff:
This looks OK code wise, modulo the issues with loading multiple changes into the same patch, but I have not done more than scan the text of the docs.

Since you said that you mostly wanted code review from me I can OK this, if you want to push as is and fix the docs later, but otherwise it is probably best to leave that until after someone else dos some review on the documentation text?

VariantValidator/modules/utils.py
Peter-J-Freeman: I'll patch this bit up. See of I can make it slicker

Peter-J-Freeman:
I want to merge into master so that the docs can build in MkDocs to make it easier for review. The its easier to just fix the docs by branching from develop. I know the docs will not be perfect first time, but they are a decent first draft. I also want a version built so that I can check links referencing the live docs, so its needed.

VariantValidator/modules/utils.py
Peter-J-Freeman: I made a compromise that implements your suggestion. We need to add a schema for error messages to contributing.md

Peter-J-Freeman: @John-F-Wagstaff . If this compromise is OK, let me know and I can merge into dev and master. Thanks. p.s. good idea on splitting the lookup tables. I hope this commit blends maintainbility and speed

John-F-Wagstaff: Looks a bit fragile for my liking, but it works, so it should be fine.  We control the origin of all the warnings here anyway.

Peter-J-Freeman: Agreed. I think a review of the Error formats and coding should be added to https://github.com/openvar/variantValidator/issues/864#issuecomment-5003082723. This would mean that over time, it would only need to recode vv_hgvs messages unless we tackle these too and update the code

ifokkema:
> I want to merge into master so that the docs can build in MkDocs to make it easier for review. The its easier to just fix the docs by branching from develop. I know the docs will not be perfect first time, but they are a decent first draft. I also want a version built so that I can check links referencing the live docs, so its needed.

I _think_ Mkdocs can be instructed to build any given branch... I'm pretty sure we've done it for the HGVS Nomenclature pages... :thinking:  (otherwise, none of this is my business haha, I'm just being nosy)

Peter-J-Freeman:
> I think Mkdocs can be instructed to build any given branch... I'm pretty sure we've done it for the HGVS Nomenclature pages... 🤔 (otherwise, none of this is my business haha, I'm just being nosy)

It possibly can be, but will run from main normally, so I really want the internal links to work. :)

PR #873, opened 2026-07-30
Moves a few regularly used dict and list creations out of functions

Peter-J-Freeman:
Also removes a stale import in vvMixinConverters.py which shows at least
 1 reverse complement func has been killed


John-F-Wagstaff (approved):
This is fairly simple and looks good as is, it should have no consequences for the logic either way.

Peter-J-Freeman:
Yep, just a quick tweak I missed. The aa dicts will speed up all calls a bit, the header tuple will speed up the batch tool mostly

Issue #876, opened 2026-07-31
Concept - Select caches in vvMixinInit


Peter-J-Freeman:
Testing a concept `SeqFetcher.fetch_seq()` LRU cache produced a modest improvement in average runtime during benchmarking, but overall execution times were more consistent. The benchmark is also likely to underestimate the benefit in production because VariantValidator commonly exhibits sequence locality.

During gap mapping of VCF files, the converter repeatedly processes the same transcript/genomic accession combinations while generating the different HGVS representations (genomic, transcript, protein and related formats). This results in repeated requests for the same sequence regions within tight processing loops, allowing an LRU cache to eliminate many duplicate sequence retrievals.

Similar locality is expected in routine clinical and research workflows. Variants are commonly analysed in batches originating from the same sequencing sample, gene panel or disease cohort, where multiple variants frequently occur within the same gene or transcript. Successive variants therefore repeatedly access the same transcript and genomic sequence, allowing cached sequence data to be reused across validations. Although the average benchmark improvement is modest, the practical benefit is expected to be greater for real-world batch processing workloads and for gap mapping, where repeated access to the same sequence regions is common.

The same principle may also be applicable to the HGVS Data Provider (HDP). Rather than introducing caching throughout the codebase, it may be possible to wrap the expensive external resources when they are created in `vvMixinInit`, allowing repeated lookups to be cached transparently while leaving the remainder of VariantValidator unchanged and not relying on cache all implementation via vvhgvs data providers 

## Current architecture

```text
                 Validator
                     │
         ┌───────────┴───────────┐
         │                       │
    SeqFetcher                 HDP
         │                       │
    fetch_seq()         get_tx_exons()
                        get_tx_identity_info()
                        get_tx_mapping_options()
                        ...
                              │
                          Database
```

## Proposed architecture

```text
                      Validator
                          │
          ┌───────────────┴───────────────┐
          │                               │
 CachedSeqFetcher                  CachedHDP
          │                               │
     SeqFetcher                         HDP
          │                               │
     fetch_seq()                 get_tx_exons()
                                 get_tx_identity_info()
                                 get_tx_mapping_options()
                                 ...
                                       │
                                   Database
```

## Validator initialisation

Current:

```python
self.sf = SeqFetcher(...)
self.hdp = hdp_connect(...)
```

Proposed:

```python
self.sf = CachedSeqFetcher(
    SeqFetcher(...)
)

self.hdp = CachedHDP(
    hdp_connect(...)
)
```

[...]

PR #877, opened 2026-07-31
Add LRU caching concept for SeqFetcher and HDP

Peter-J-Freeman:
Implement in-process LRU cache decorators for the SeqFetcher and HGVS data provider (HDP) to reduce repeated lookups during validation.

The SeqFetcher cache stores repeated fetch_seq() requests. The HDP cache stores results from get_tx_identity_info(), get_tx_for_gene(), get_pro_ac_for_tx_ac(), get_tx_exons(), get_gene_info() and get_tx_mapping_options(). get_tx_for_region() was intentionally excluded because these genomic coordinate queries are unlikely to repeat during validation.

Both wrappers use the decorator pattern and transparently delegate all uncached methods to the underlying implementations.

Benchmarking shows the SeqFetcher cache primarily improves runtime stability, while the addition of the HDP cache provides a substantial reduction in overall validation time.

Refs #876

### DIFF
--- a/VariantValidator/modules/vvMixinInit.py
+++ b/VariantValidator/modules/vvMixinInit.py
@@ -288,6 +288,14 @@ class Mixin:
         """
 
         # --------------------------------------------------------------
+        # HGVS global configuration
+        # --------------------------------------------------------------
+
+        vvhgvs.global_config.uta.pool_max = 25
+        vvhgvs.global_config.formatting.max_ref_length = 1000000
+        vvhgvs.global_config.lru_cache.maxsize = 1000
+
+        # --------------------------------------------------------------
         # Configuration
         # --------------------------------------------------------------
 
@@ -437,13 +445,6 @@ class Mixin:
         self.no_norm_evm = None
 
         # --------------------------------------------------------------
-        # HGVS global configuration
-        # --------------------------------------------------------------
-
-        vvhgvs.global_config.uta.pool_max = 25
-        vvhgvs.global_config.formatting.max_ref_length = 1000000
-
-        # --------------------------------------------------------------
         # HGVS data provider
         # --------------------------------------------------------------
 
@@ -457,7 +458,7 @@ class Mixin:
         # transparently to the original HGVS data provider via __getattr__().
         #
         # To disable all HDP caching, simply comment out the line below.
-        self.hdp = CachedHDP(self.hdp)
+        # self.hdp = CachedHDP(self.hdp)
 
         self.utaSchema = str(
             self.hdp.data_version()
--- a/VariantValidator/modules/vvMixinInit.py
+++ b/VariantValidator/modules/vvMixinInit.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+from functools import lru_cache
 from configparser import ConfigParser
 
 import vvhgvs
@@ -40,6 +41,216 @@ class InitialisationError(Exception):
     """Raised when the Validator cannot be initialised."""
     pass
 
+
+class CachedSeqFetcher:
+    """
+    LRU cache wrapper for SeqFetcher.
+
+    This class implements the Decorator pattern around SeqFetcher.
+    Only the most frequently repeated sequence lookup is cached using
+    functools.lru_cache(); all other methods and attributes are
+    transparently delegated to the wrapped SeqFetcher.
+
+    The cache exists solely within the current Python process and is
+    discarded when the process exits.
+    """
+
+    def __init__(self, sf):
+        """
+        Wrap an existing SeqFetcher.
+
+        Parameters
+        ----------
+        sf
+            An instantiated SeqFetcher object.
+        """
+        self._sf = sf
+
+    def __getattr__(self, name):
+        """
+        Delegate all uncached methods and attributes to the wrapped
+        SeqFetcher.
+
+        Python only calls __getattr__ when an attribute is not found on
+        CachedSeqFetcher itself. Consequently, only the methods
+        explicitly implemented below are intercepted and cached;
+        everything else behaves exactly as if the original SeqFetcher
+        were being used directly.
+        """
+        return getattr(self._sf, name)
+
+    @lru_cache(maxsize=32768)
+    def fetch_seq(self, ac, start_i=None, end_i=None):
+        """
+        Retrieve a sequence or sequence slice.
+
+        The cache key is formed from the accession, start coordinate
+        and end coordinate. Repeated requests for the same sequence
+        slice are therefore served directly from memory rather than
+        performing another SeqRepo lookup.
+        """
+        return self._sf.fetch_seq(ac, start_i, end_i)
+
+    def clear_caches(self):
+        """
+        Clear every LRU cache maintained by this wrapper.
+
+        This is primarily intended for:
+            * benchmarking cold versus warm cache performance;
+            * releasing cached sequence slices;
+            * forcing fresh sequence lookups.
+        """
+        self.fetch_seq.cache_clear()
+
+    def cache_info(self):
+        """
+        Return cache statistics for every cached method.
+
+        Each entry contains the standard functools CacheInfo tuple:
+            hits
+            misses
+            maxsize
+            currsize
+
+        This method is intended for benchmarking and determining
+        whether the cache provides sufficient benefit to justify its
+        memory usage.
+        """
+        return {
+            "fetch_seq": self.fetch_seq.cache_info(),
+        }
+
+
+class CachedHDP:
+    """
+    LRU cache wrapper for the HGVS data provider.
+
+    This class implements the Decorator pattern around the HGVS UTA data
+    provider. Only the most frequently repeated database lookups are
+    cached using functools.lru_cache(); all other methods and attributes
+    are transparently delegated to the wrapped provider.
+
+    The cache exists solely within the current Python process and is
+    discarded when the process exits.
+    """
+
+    def __init__(self, hdp):
+        """
+        Wrap an existing HGVS data provider.
+
+        Parameters
+        ----------
+        hdp
+            An instantiated HGVS data provider returned by
+            vvhgvs.dataproviders.uta.connect().
+        """
+        self._hdp = hdp
+
+    def __getattr__(self, name):
+        """
+        Delegate all uncached methods and attributes to the wrapped
+        HGVS data provider.
+
+        Python only calls __getattr__ when an attribute is not found on
+        CachedHDP itself. Consequently, only the methods explicitly
+        implemented below are intercepted and cached; everything else
+        behaves exactly as if the original HGVS data provider were being
+        used directly.
+        """
+        return getattr(self._hdp, name)
+
+    @lru_cache(maxsize=8192)
+    def get_tx_identity_info(self, tx_ac):
+        """
+        Retrieve transcript identity information for a transcript
+        accession.
+
+        This is the most frequently repeated HDP lookup performed by
+        VariantValidator and therefore has the largest cache.
+        """
+        return self._hdp.get_tx_identity_info(tx_ac)
+
+    @lru_cache(maxsize=4096)
+    def get_tx_for_gene(self, gene):
+        """
+        Retrieve all mapped transcripts associated with an HGNC gene
+        symbol.
+        """
+        return self._hdp.get_tx_for_gene(gene)
+
+    @lru_cache(maxsize=4096)
+    def get_pro_ac_for_tx_ac(self, tx_ac):
+        """
+        Retrieve the associated protein accession for a transcript
+        accession.
+        """
+        return self._hdp.get_pro_ac_for_tx_ac(tx_ac)
+
+    @lru_cache(maxsize=4096)
+    def get_tx_exons(self, tx_ac, alt_ac, alt_aln_method):
+        """
+        Retrieve transcript exon structures for a transcript/genome
+        alignment.
+        """
+        return self._hdp.get_tx_exons(tx_ac, alt_ac, alt_aln_method)
+
+    @lru_cache(maxsize=2048)
+    def get_gene_info(self, gene):
+        """
+        Retrieve HGNC gene information for a gene symbol.
+        """
+        return self._hdp.get_gene_info(gene)
+
+    @lru_cache(maxsize=4096)
+    def get_tx_mapping_options(self, tx_ac, gap_warn=False):
+        """
+        Retrieve all genomic mapping options for a transcript.
+
+        The gap_warn argument forms part of the cache key, meaning
+        cached results for gap_warn=True and gap_warn=False remain
+        independent.
+        """
+        return self._hdp.get_tx_mapping_options(tx_ac, gap_warn)
+
+    def clear_caches(self):
+        """
+        Clear every LRU cache maintained by this wrapper.
+
+        This is primarily intended for:
+            * benchmarking cold versus warm cache performance;
+            * releasing cached objects;
+            * forcing fresh database lookups after updating the
+              underlying transcript database.
+        """
+        self.get_tx_identity_info.cache_clear()
+        self.get_tx_for_gene.cache_clear()
+        self.get_pro_ac_for_tx_ac.cache_clear()
+        self.get_tx_exons.cache_clear()
+        self.get_gene_info.cache_clear()
+        self.get_tx_mapping_options.cache_clear()
+
+    def cache_info(self):
+        """
+        Return cache statistics for every cached method.
+
+        Each entry contains the standard functools CacheInfo tuple:
+            hits
+            misses
+            maxsize
+            currsize
+
+        This method is intended for benchmarking and determining
+        whether each cache provides sufficient benefit to justify its
+        memory usage.
+        """
+        return {
+            "get_tx_identity_info": self.get_tx_identity_info.cache_info(),
+            "get_tx_for_gene": self.get_tx_for_gene.cache_info(),
+            "get_pro_ac_for_tx_ac": self.get_pro_ac_for_tx_ac.cache_info(),
+            "get_tx_exons": self.get_tx_exons.cache_info(),
+            "get_gene_info": self.get_gene_info.cache_info(),
+            "get_tx_mapping_options": self.get_tx_mapping_options.cache_info(),
+        }
 
 class Mixin:
     """
@@ -236,9 +447,17 @@ class Mixin:
         # HGVS data provider
         # --------------------------------------------------------------
 
-        self.hdp = vvhgvs.dataproviders.uta.connect(
-            pooling=True,
-        )
+        # Create the HGVS data provider.
+        self.hdp = vvhgvs.dataproviders.uta.connect(pooling=True)
+
+        # Wrap the HGVS data provider with the LRU cache layer.
+        #
+        # CachedHDP intercepts only the high-frequency database lookups and
+        # caches their results. All other methods and attributes are delegated
+        # transparently to the original HGVS data provider via __getattr__().
+        #
+        # To disable all HDP caching, simply comment out the line below.
+        self.hdp = CachedHDP(self.hdp)
 
         self.utaSchema = str(
             self.hdp.data_version()
@@ -282,6 +501,15 @@ class Mixin:
         self.sf = vvhgvs.dataproviders.seqfetcher.SeqFetcher(
             self.check_same_thread,
         )
+
+        # Wrap the SeqFetcher with the LRU cache layer.
+        #
+        # CachedSeqFetcher intercepts sequence fetches and caches the returned
+        # sequence slices. All other methods and attributes are transparently
+        # delegated to the original SeqFetcher via __getattr__().
+        #
+        # To disable all SeqFetcher caching, simply comment out the line below.
+        self.sf = CachedSeqFetcher(self.sf)
 
         # --------------------------------------------------------------
         # Persistent alignment-specific normalizers


Peter-J-Freeman:
@John-F-Wagstaff. This is a concept, so feel free not to accept. I know you are careful with caches.

For reference, taken from the issues link

<html><head></head><body><h2>Benchmark methodology</h2><p>To evaluate the impact of introducing in-process LRU caches, the full<br>VariantValidator test suite was executed three times for each caching<br>configuration using:</p><pre><code class="language-bash">pytest -n 4
</code></pre><p>Each benchmark consisted of <strong>2,197 passing tests and 6 skipped tests</strong>.<br>The same hardware, Python environment and xdist configuration were used<br>throughout. Three independent runs were performed for each<br>configuration, and the mean runtime, standard deviation (SD) and<br>coefficient of variation (CV) were calculated.</p><p>Three configurations were compared:</p><ol><li><p><strong>No caching (baseline).</strong></p></li><li><p><strong>SeqFetcher cache only</strong>, caching repeated <code inline="">fetch_seq()</code> requests.</p></li><li><p><strong>SeqFetcher + HGVS Data Provider (HDP) cache.</strong></p></li></ol><p>The HDP cache stores the results of repeated calls to:</p><ul><li><p><code inline="">get_tx_identity_info()</code></p></li><li><p><code inline="">get_tx_for_gene()</code></p></li><li><p><code inline="">get_pro_ac_for_tx_ac()</code></p></li><li><p><code inline="">get_tx_exons()</code></p></li><li><p><code inline="">get_gene_info()</code></p></li><li><p><code inline="">get_tx_mapping_options()</code></p></li></ul><p><code inline="">get_tx_for_region()</code> was intentionally excluded because genomic<br>coordinate queries are expected to have a very low cache hit rate during<br>normal validation workflows.</p><h2>Results</h2>

Configuration | Run 1 (s) | Run 2 (s) | Run 3 (s) | Mean (s) | SD (s) | CV (%) | Δ Mean | Δ SD
-- | -- | -- | -- | -- | -- | -- | -- | --
No cache | 468.55 | 450.96 | 472.48 | 464.00 | 11.46 | 2.47 | – | –
SeqFetcher cache | 463.84 | 459.06 | 459.62 | 460.84 | 2.61 | 0.57 | -0.68% | -77.2%
SeqFetcher + HDP cache | 412.18 | 435.23 | 415.53 | 420.98 | 12.56 | 2.98 | -9.27% | +9.6%

<h2>Conclusions</h2><p>Caching repeated sequence retrievals alone provides only a modest<br>improvement in overall runtime (0.68%), but it substantially reduces<br>run-to-run variability, decreasing the standard deviation by<br>approximately 77%. This indicates that sequence retrieval is not a major<br>runtime bottleneck but that eliminating repeated fetches makes execution<br>more consistent.</p><p>The majority of the performance improvement comes from caching HGVS Data<br>Provider lookups. When the HDP cache is combined with the SeqFetcher<br>cache, the mean runtime falls from <strong>464.00 seconds</strong> to <strong>420.98<br>seconds</strong>, representing a reduction of approximately <strong>43 seconds</strong><br>(<strong>9.27%</strong>).</p><p>These results suggest that repeated transcript and annotation lookups<br>are a significant component of VariantValidator execution time.<br>Specifically, caching transcript identity, transcript-to-gene<br>relationships, transcript exon structures, transcript mapping options,<br>protein accessions and gene information removes a substantial amount of<br>repeated database work during validation.</p><p>Although the combined cache exhibits similar run-to-run variability to<br>the uncached implementation, the overall reduction in execution time is<br>large enough to clearly demonstrate that the HGVS Data Provider cache is<br>responsible for the vast majority of the observed performance gains,<br>with the SeqFetcher cache providing a smaller complementary benefit by<br>avoiding repeated sequence retrievals.</p></body></html>

https://github.com/openvar/variantValidator/issues/876

John-F-Wagstaff:
Everything mentioned here, including `get_seq` already goes through

https://github.com/openvar/vv_hgvs/blob/master/vvhgvs/dataproviders/interface.py

Which includes an lru cache for all of this.

Yet despite this the benefit of a bit more hits on the cash is enough to make this look like a performance gain, especially for SeqRepo. 

I currently have no idea what would help for SeqRepo usage, aside from bumping the global cash size.

I added TranscriptMapData to help this issue (repeated re-calls of expensive DB fetches), it caches the first fetch it gets for each transcript/transcript-genomic ref pair, for the lifetime of the validation. This works on a per variant basis at the moment.  It also currently only affects some `get_tx_exons` and `get_tx_mapping_options` but should probably be expanded to all, along with some other queries, suhch as  `get_tx_identity_info` and possibly some others. 

I have no idea how much expanding this would help in total VS creating an extra layer of local caching like this PR does, or upping the global cache size however, it did speed things up somewhat when I was testing it, even in circumstances that the global cash **should** have invalidated the gain.

Peter-J-Freeman:
So the conculusion is `well it works, but we aren't fully sure why` :)
Well, do we argue it or just go with it?

I am happy to change the cache sizes etc to see if I can make it even faster. 

What do you think?

Peter-J-Freeman:
There will be a global removal of a SeqRepo call when a project to map Selenon genes properly concludes in a few weeks and we do the next database build. This will help a lot

John-F-Wagstaff:
Not quite, I have suspicions of why, namely that if you do tests in parallel you might get enough fetches to overflow the cache between "related" calls, causing a re-fetch, the cache size is fixed. 

When I said the **global** cache size I was talking about the cache in vv_hgvs, if this cache size is increased it should affect the validation +replace ref calls internal to vvhgvs too.  If we get the same performance boost from that then we don't want to add another layer of caching.

John-F-Wagstaff:
We also might see some changes to performance between using `fetch_seq` from the vvhgvs and pulling out the vvhgvs seq fetcher and then using it separate from that (i.e. we might be unintentionally end running around the cache with the way we use the seq fetcher _VS_ doing hdp.seq_fech()). 

Adding an extra layer of caching and a load of extra code to maintain to undo that if it is what is happening would be a waste.

Peter-J-Freeman:
cool so its an in vvta and seqrepo thing. But this shows its worth investigating so was not useless :)

John-F-Wagstaff:
Who knows what happens if we bump the vvhgvs cache size though? it is probably worth checking.  

The entire original vvhgvs code base was architected by someone who side-loaded a cache of correct answers in before running the tests! He did do some complex checks depending on cache state, but in the state we forked from it effectively defaulted to just testing the existing "pickled" cache unless steps were taken to prevent this. (and he also thought, or possibly still thinks, that even heavy users should fetch from remote servers like everyone else, but just cache harder, from the way thing are written at least. )

If you do want to do a test we need to change 
`hgvs.global_config.lru_cache.maxsize`
either that or edit 
`vv_hgvs/vvhgvs/_data/defaults.ini`
locally while running the tests. The original default is 100 distinct queries (per cached function I believe) so upping it to 200 should show some affect, if we get no affect then that is a result, but a low but real affect probably means that we are end-running around the cache for seqrepo but the rest is working.

Peter-J-Freeman:
I prefer to do it in vvminininit to make it easier to maintain

```
        # --------------------------------------------------------------
        # HGVS global configuration
        # --------------------------------------------------------------

        vvhgvs.global_config.uta.pool_max = 25
        vvhgvs.global_config.formatting.max_ref_length = 1000000
```        

Setting to 200. Do you want me to delete my local caches or keep them and see what happens, or both?

John-F-Wagstaff:
I don't know what will happen if you change the settings, I don't know whether adding the extra lru cache on top will have any affect at all on performance with an increased underlying cache size. 

So we have to try the combinations,
- boosted underlying cache no overlying cache
- boosted underlying cache seqrepo overlying cache
- boosted underlying cache full overlying cache
and see what happens. Then we can think about what that means. We might want to try changing the cache quantity too e.g. test 200 vs 400, (vs implicit test of 100, which is already tested as the default) but that depends on the result.

Peter-J-Freeman:
cool can do all this. By the way, pytest or pytest -n 4 for best results that are more realistic?

John-F-Wagstaff:
whatever you did for your first tests for now, I think, so that we don't have to re-do a baseline.

Peter-J-Freeman:
I'm running this now. Just got started, but the results are shocking. Will publish when done

Peter-J-Freeman:
OK, this is really unexpected

<h2 data-section-id="1leblw4" data-start="223" data-end="267" class="PDq2pG_selectionAnchorContainer">Final benchmark table (slowest → fastest)<span aria-hidden="true" class="PDq2pG_selectionAnchor"></span></h2>
<div class="TyagGW_tableContainer"><div tabindex="-1" class="group TyagGW_tableWrapper flex flex-col-reverse w-fit">

Configuration | Run 1 (s) | Run 2 (s) | Run 3 (s) | Mean (s) | SD (s) | CV (%) | Δ Mean | Δ SD
-- | -- | -- | -- | -- | -- | -- | -- | --
No cache (HGVS default) | 468.55 | 450.96 | 472.48 | 464.00 | 11.46 | 2.47 | – | –
SeqFetcher cache (HGVS default) | 463.84 | 459.06 | 459.62 | 460.84 | 2.61 | 0.57 | -0.68% | -77.2%
SeqFetcher + HDP cache (HGVS default) | 412.18 | 435.23 | 415.53 | 420.98 | 12.56 | 2.98 | -9.27% | +9.6%
SeqFetcher + HDP cache + HGVS LRU = 200 | 337.47 | 337.45 | 344.44 | 339.79 | 4.03 | 1.19 | -26.77% | -64.8%
SeqFetcher + HGVS LRU = 200 | 335.85 | 337.58 | 339.56 | 337.66 | 1.86 | 0.55 | -27.23% | -83.8%
SeqFetcher + HGVS LRU = 400 | 312.71 | 300.03 | 296.56 | 303.10 | 8.42 | 2.78 | -34.68% | -26.5%
SeqFetcher + HDP cache + HGVS LRU = 400 | 287.05 | 285.29 | 301.05 | 291.13 | 8.76 | 3.01 | -37.26% | -23.6%
SeqFetcher + HDP cache + HGVS LRU = 600 | 272.08 | 275.53 | 267.82 | 271.81 | 3.86 | 1.42 | -41.42% | -66.3%
SeqFetcher + HDP cache + HGVS LRU = 800 | 262.48 | 265.58 | 268.16 | 265.41 | 2.86 | 1.08 | -42.80% | -75.0%
SeqFetcher + HDP cache + HGVS LRU = 1000 | 251.92 | 254.46 | 254.83 | 253.74 | 1.59 | 0.63 | -45.31% | -86.1%

</div></div>
<h2 data-section-id="1h1j6g0" data-start="1611" data-end="1639">T</h2>

Caching clearly helps, but whats the sweet spot?

John-F-Wagstaff:
What are the results without the SeqFetcher cache ?

We specifically need to know what happens with the vv_hgvs lru cache goes up without any VV local caching VS with or we can't tell wether it is worth adding the VV local caching. 

The gain from adding the local cache VS running without the local cache should drop as the vv_hgvs lru cache goes up, but how much?

We need:
- No VV cache, HGVS LRU = 200
- No VV cache, HGVS LRU = 400
and possibly
- No VV cache, HGVS LRU = 1000

if the difference after a certain point is less than a couple of seconds out of more than 200 that changes things.

Peter-J-Freeman:
Yep, I ran out of time to do the runs with the local hdp cache on, and the local seqfetcher cache on, then both off. Will add these over the weekend. I decided to try max out the gain first to get a ball park, but didn't manage to, but it's all informative and will help get to a conclusion.

Peter-J-Freeman:
<h2 data-section-id="1a521wz" data-start="0" data-end="53" class="PDq2pG_selectionAnchorContainer">Table 2 – Follow-up benchmarking (evening session)<span aria-hidden="true" class="PDq2pG_selectionAnchor"></span></h2>
<div class="TyagGW_tableContainer"><div tabindex="-1" class="group TyagGW_tableWrapper flex flex-col-reverse w-fit">

Configuration | Run 1 (s) | Run 2 (s) | Run 3 (s) | Mean (s) | SD (s) | CV (%) | Δ Mean
-- | -- | -- | -- | -- | -- | -- | --
Local HDP cache OFF, SeqFetcher cache OFF, HGVS LRU = 1000 | 330.31 | 330.35 | 332.62 | 331.09 | 1.30 | 0.39 | -28.64%
SeqFetcher + HGVS LRU = 600 | 330.94 | 321.27 | 302.76 | 318.32 | 14.32 | 4.50 | -31.40%
SeqFetcher + Local HDP cache + HGVS LRU = 1000 | 282.75 | 286.72 | 285.78 | 285.08 | 2.08 | 0.73 | -38.56%
SeqFetcher + Local HDP cache + HGVS LRU = 1200 | 279.61 | 283.80 | 285.39 | 282.93 | 2.99 | 1.06 | -39.02%
SeqFetcher + HGVS LRU = 1000 | 286.94 | 278.94 | 279.10 | 281.66 | 4.57 | 1.62 | -39.30%
SeqFetcher + HGVS LRU = 1200 | 279.38 | 276.04 | 277.56 | 277.66 | 1.67 | 0.60 | -40.16%

</div></div>
<h3 data-section-id="wv8cei" data-start="925" data-end="936">Summary</h3>
<p data-start="938" data-end="1281">The second benchmarking session investigated the contribution of the <strong data-start="1007" data-end="1034">HGVS internal LRU cache</strong> independently of the custom <strong data-start="1063" data-end="1076">Local HDP</strong> and <strong data-start="1081" data-end="1095">SeqFetcher</strong> wrapper caches. Increasing the HGVS internal cache continued to improve performance from <strong data-start="1185" data-end="1192">600</strong> to <strong data-start="1196" data-end="1204">1000</strong> entries, with a smaller additional improvement observed at <strong data-start="1264" data-end="1272">1200</strong> entries.</p>
<p data-start="1283" data-end="1744">Disabling both the <strong data-start="1302" data-end="1315">Local HDP</strong> and <strong data-start="1320" data-end="1334">SeqFetcher</strong> caches while retaining an HGVS cache size of <strong data-start="1380" data-end="1388">1000</strong> increased the mean runtime to <strong data-start="1419" data-end="1431">331.09 s</strong>, demonstrating that the wrapper caches contribute additional performance improvements. However, comparison of the remaining configurations indicates that once the HGVS internal cache is increased to <strong data-start="1631" data-end="1644">1000–1200</strong> entries, the majority of the performance benefit is attributable to the HGVS internal cache itself.</p>
<p data-start="1746" data-end="1807">Within this benchmark session, the fastest configuration was:</p>
<ul data-start="1809" data-end="1907">
<li data-section-id="hhc3nx" data-start="1809" data-end="1839">
<strong data-start="1811" data-end="1839">SeqFetcher cache enabled</strong>
</li>
<li data-section-id="1rrzox1" data-start="1840" data-end="1870">
<strong data-start="1842" data-end="1870">Local HDP cache disabled</strong>
</li>
<li data-section-id="dcoyyk" data-start="1871" data-end="1907">
<strong data-start="1873" data-end="1907">HGVS internal LRU cache = 1200</strong>
</li>
</ul>
<p data-start="1909" data-end="2051">This achieved a mean runtime of <strong data-start="1941" data-end="1953">277.66 s</strong>, corresponding to a <strong data-start="1974" data-end="1994">40.16% reduction</strong> relative to the uncached baseline measured in Session 1.</p>
<p data-start="2053" data-end="2432">The difference between <strong data-start="2076" data-end="2095">HGVS LRU = 1000</strong> and <strong data-start="2100" data-end="2119">HGVS LRU = 1200</strong> was modest (<strong data-start="2132" data-end="2144">281.66 s</strong> versus <strong data-start="2152" data-end="2164">277.66 s</strong>), indicating diminishing returns as the cache size increases. Based on these data, a cache size of approximately <strong data-start="2278" data-end="2322">1000–1200 entries per cached HGVS method</strong> appears to capture most of the achievable performance benefit for the VariantValidator functional test suite.</p>
<p data-start="2434" data-end="2866" data-is-last-node="" data-is-only-node="">The benchmark also suggests that the additional <strong data-start="2482" data-end="2495">Local HDP</strong> wrapper cache provides little or no measurable benefit once the HGVS internal cache has been increased to this range.</p>

John-F-Wagstaff:
So if we intend to bump the cache to 1.2k or so we **just** want to boost the vvhgvs cache setting, and not pull this?

I should note that since the tests end up repeatedly target the same problem transcripts in different ways, we can not expect the same level of performance gain for user input as we get for tests (from either cache version), though some specific batch tests may gain a outsized benefit.

Peter-J-Freeman:
<h2 data-section-id="1yj6rxr" data-start="0" data-end="76" class="PDq2pG_selectionAnchorContainer">SeqFetcher cache benchmarking (HGVS LRU = 1000, Local HDP cache disabled)<span aria-hidden="true" class="PDq2pG_selectionAnchor"></span></h2>
<div class="TyagGW_tableContainer"><div tabindex="-1" class="group TyagGW_tableWrapper flex flex-col-reverse w-fit">

SeqFetcher cache size | Mean (s) | SD (s) | CV (%) | Δ Mean
-- | -- | -- | -- | --
OFF | 307.39 | 2.99 | 1.00 | –
4096 | 261.71 | 0.91 | 0.36 | -14.86%
8192 | 260.30 | 1.39 | 0.55 | -15.31%
16384 | 257.87 | 1.52 | 0.61 | -16.11%
32768 | 252.50 | 1.32 | 0.52 | -17.86%
65536 | 249.68 | 2.83 | 1.13 | -18.78%

</div></div>
<h3 data-section-id="wv8cei" data-start="500" data-end="511">Summary</h3>
<p data-start="513" data-end="879">The final phase of benchmarking evaluated the impact of the SeqFetcher LRU cache while keeping the HGVS internal LRU cache fixed at <strong data-start="645" data-end="661">1000 entries</strong> and the Local HDP cache disabled. Enabling the SeqFetcher cache immediately produced a substantial improvement in execution time, reducing the runtime by approximately <strong data-start="830" data-end="837">15%</strong> compared with having no SeqFetcher cache.</p>
<p data-start="881" data-end="1288">Increasing the cache size beyond <strong data-start="914" data-end="922">4096</strong> continued to improve performance, although the magnitude of the improvement steadily decreased, demonstrating the expected diminishing returns of an LRU cache. Performance continued to improve through <strong data-start="1124" data-end="1132">8192</strong>, <strong data-start="1134" data-end="1143">16384</strong>, <strong data-start="1145" data-end="1154">32768</strong>, and <strong data-start="1160" data-end="1169">65536</strong> entries, with the largest gains achieved at the smaller cache sizes and progressively smaller improvements thereafter.</p>
<p data-start="1290" data-end="1647">The fastest configuration tested was <strong data-start="1327" data-end="1336">65536</strong> entries, achieving an <strong data-start="1359" data-end="1369">18.78%</strong> reduction in runtime relative to the uncached configuration. However, the improvement over <strong data-start="1461" data-end="1470">32768</strong> entries was relatively modest (approximately <strong data-start="1516" data-end="1525">2.8 s</strong>, or <strong data-start="1530" data-end="1542">0.9–1.1%</strong>), suggesting that the cache is approaching saturation for the VariantValidator functional test workload.</p>
<p data-start="1649" data-end="2046" data-is-last-node="" data-is-only-node="">Overall, these benchmarks indicate that the SeqFetcher cache provides a significant performance benefit, while increasing the cache beyond <strong data-start="1788" data-end="1797">32768</strong> entries yields only marginal additional improvement. Consequently, <strong data-start="1865" data-end="1882">32768 entries</strong> remains a sensible default configuration, providing near-optimal performance while avoiding unnecessary growth in cache size.</p>


Agreed. We should not pull this. I will make a fresh cleaned branch, but wanted to add this in before doing it. I think the sf cache is optimal already too

Peter-J-Freeman:
<div class="" data-turn-id-container="d93c7e3b-b714-4c82-b553-7053290fc436" data-is-intersecting="true"><section class="text-token-text-primary w-full focus:outline-none has-data-writing-block:pointer-events-none [&amp;:has([data-writing-block])&gt;*]:pointer-events-auto R6Vx5W_threadScrollVars scroll-mb-[calc(var(--scroll-root-safe-area-inset-bottom,0px)+var(--thread-response-height))] scroll-mt-(--sticky-padding-top)" dir="auto" data-turn-id="d93c7e3b-b714-4c82-b553-7053290fc436" data-turn-id-container="d93c7e3b-b714-4c82-b553-7053290fc436" data-testid="conversation-turn-707" data-turn="user"></section></div><div class="" data-turn-id-container="request-6a6b453c-7d00-83eb-a2d7-aff781170690-537" data-is-intersecting="true"><section class="text-token-text-primary w-full focus:outline-none has-data-writing-block:pointer-events-none [&amp;:has([data-writing-block])&gt;*]:pointer-events-auto R6Vx5W_threadScrollVars scroll-mb-[calc(var(--scroll-root-safe-area-inset-bottom,0px)+var(--thread-response-height))] scroll-mt-[calc(var(--header-height)+min(200px,max(70px,20svh)))]" dir="auto" data-turn-id="request-6a6b453c-7d00-83eb-a2d7-aff781170690-537" data-turn-id-container="request-6a6b453c-7d00-83eb-a2d7-aff781170690-537" data-testid="conversation-turn-708" data-turn="assistant"><div class="text-base my-auto mx-auto [--thread-content-margin:var(--thread-content-margin-xs,calc(var(--spacing)*4))] @w-sm/main:[--thread-content-margin:var(--thread-content-margin-sm,calc(var(--spacing)*6))] @w-lg/main:[--thread-content-margin:var(--thread-content-margin-lg,calc(var(--spacing)*16))] px-(--thread-content-margin)"><div data-conversation-screenshot-content="" class="[--thread-content-max-width:40rem] @w-lg/main:[--thread-content-max-width:48rem] mx-auto max-w-(--thread-content-max-width) flex-1 group/turn-messages focus-visible:outline-hidden relative flex w-full min-w-0 flex-col agent-turn"><div class="flex max-w-full flex-col gap-4 grow"><div data-message-author-role="assistant" data-message-id="c115501d-41a8-405c-a46b-5cee907d4c94" dir="auto" data-message-model-slug="gpt-5-5" class="min-h-8 text-message relative flex w-full flex-col items-end gap-2 text-start break-words whitespace-normal outline-none keyboard-focused:focus-ring [.text-message+&amp;]:mt-1"><div class="flex w-full flex-col gap-1 empty:hidden"><div class="markdown prose dark:prose-invert wrap-break-word w-full light markdown-new-styling"><div class="TyagGW_tableContainer"><div tabindex="-1" class="group TyagGW_tableWrapper flex flex-col-reverse w-fit">

Rank | Local HDP Cache | SeqFetcher Cache | HGVS Global Cache | DBGet Cache | Mean (s) | Mean (min) | Δ Mean
-- | -- | -- | -- | -- | -- | -- | --
1 (Baseline) | Off | Off | Off | Off | 1594.66 | 26.58 | Baseline
2 | Off | On (32768) | 1000 | Off | 280.78 | 4.68 | −1313.88 s (−82.4%)
3 | Off | On (32768) | 1000 | On (10000) | 259.15 | 4.32 | −1335.51 s (−83.8%)
4 (Best) | Off | On (32768) | 1000 | On (15000) | 251.60 | 4.19 | −1343.06 s (−84.2%)

</div></div>
<h3 data-section-id="wv8cei" data-start="597" data-end="608">Summary</h3>
<p data-start="610" data-end="1039">The final benchmarking demonstrates that the combined caching strategy provides a substantial improvement in VariantValidator performance. With all caches disabled, the complete test suite required an average of <strong data-start="822" data-end="839">26.58 minutes</strong>. Enabling the recommended caching configuration reduced this to <strong data-start="904" data-end="920">4.19 minutes</strong>, representing an overall reduction of <strong data-start="959" data-end="994">1343.06 seconds (22.38 minutes)</strong>, or an <strong data-start="1002" data-end="1038">84.2% decrease in execution time</strong>.</p>
<p data-start="1041" data-end="1090">The recommended cache configuration is therefore:</p>
<ul data-start="1092" data-end="1260">
<li data-section-id="1dmgyc5" data-start="1092" data-end="1128">
<strong data-start="1094" data-end="1119">Local HGVS HDP cache:</strong> Disabled
</li>
<li data-section-id="1k4jj0w" data-start="1129" data-end="1177">
<strong data-start="1131" data-end="1152">SeqFetcher cache:</strong> Enabled (32,768 entries)
</li>
<li data-section-id="r4k6w0" data-start="1178" data-end="1216">
<strong data-start="1180" data-end="1202">HGVS global cache:</strong> 1,000 entries
</li>
<li data-section-id="l6gj0g" data-start="1217" data-end="1260">
<strong data-start="1219" data-end="1235">DBGet cache:</strong> Enabled (15,000 entries)
</li>
</ul>
<p data-start="1262" data-end="1481" data-is-last-node="" data-is-only-node="">Further increases in cache size did not produce measurable improvements, indicating that these settings provide a good balance between execution speed and memory consumption for the VariantValidator validation workload.</p></div></div></div></div></div></div></section></div>

Peter-J-Freeman:
> I should note that since the tests end up repeatedly target the same problem transcripts in different ways, we can not expect the same level of performance gain for user input as we get for tests (from either cache version), though some specific batch tests may gain a outsized benefit.

I agree. The test suite is something of a best-case scenario because it repeatedly validates the same genes and transcripts, allowing the caches to warm up and be reused. We therefore should not expect the same magnitude of improvement for a single user validation.

That said, I think there are several important benefits:

Individual validations should still improve because a single validation performs multiple repeated SeqFetcher, database and VVTA/HDP lookups internally, so there are opportunities for cache hits even within a single request.
Multi-worker execution also benefits. The benchmarking was performed using four pytest workers, demonstrating that the caching strategy remains effective under concurrent workloads. This is directly applicable to the REST API, where multiple worker processes are servicing requests simultaneously.
The REST API and batch processing are likely to benefit the most. These workflows frequently process many variants from the same gene or transcript, particularly when all-transcript or raw transcript reporting is enabled. Those workloads naturally generate repeated lookups, making them ideal candidates for this caching strategy.
The optimisation specifically targets some of the most expensive operations in the validation pipeline—database lookups, SeqRepo sequence retrieval and HGVS data provider access. Reducing repeated I/O is likely to provide more benefit than attempting to optimise code that is already computationally efficient.
The remaining major performance bottleneck is likely to be normalisation. Unlike the lookup stages, this is largely computational rather than I/O bound, so it will probably require a different optimisation strategy than caching.

Finally, one aspect I think is important is that this work was benchmark-driven rather than assumption-driven. We didn't simply add caches everywhere. Each cache was benchmarked, cache sizes were tuned empirically, and caches that showed little or no measurable benefit were either reduced to their optimum size or deliberately not implemented. The result is a caching strategy that balances execution speed, memory usage and maintainability, rather than simply maximising cache size or caching every possible lookup.

Peter-J-Freeman:
An additional point worth noting is that the comparison against the last develop branch is conservative. Since that point, the project has gained a substantial number of new unit and regression tests, increasing the overall workload of the CI pipeline. Despite this, the end-to-end GitHub Actions runtime has reduced from approximately 36 minutes on develop to approximately 22 minutes on the final optimisation branch. This represents an improvement of almost 14 minutes (≈38%), even though the pipeline is now executing a larger and more comprehensive test suite.

In other words, the optimisation has not simply made the previous workload faster—it has made a larger workload complete in significantly less time. That improvement is the cumulative result of the codebase clean-up, removal of unnecessary regex and string processing, increased use of HGVS objects, reduced object creation, improved lookup paths, targeted caching, and benchmark-driven cache tuning. Together, these changes have improved maintainability, increased automated test coverage, and substantially reduced execution time for both local development and continuous integration.

I strongly belive that when you add your own optimisations @John-F-Wagstaff h, i.e. the fine-tooth-comb clean, we have done a really good job of this refactor. Also, test coverage is now just marginally below 85%.

John-F-Wagstaff:
The very fact that the seqFetcher cache produces performance results like this however means that we are probably doing an end-run around the hdp inbuilt caching for sequence fetch. Otherwise the difference between no caching and the first cached test should probably be a lot smaller, though I would be more certain if you had set the cache size to 1000 for the first test not 4 times that. 

This does allow us to have a higher seq fetch cache than the for the other endpoints though. Despite this should probably see a better results for the same cache size if the vvhgvs cache is also used by the VV, so in future we may want to add the provision to have a different vvhgvs internal cache size for seq fetch than the other endpoints, fix the accidental end run, and then remove the extra seq fetcher cache code from VV to avoid the unneeded extra complexity.

PR #879, opened 2026-08-02
Cache optimisation

Peter-J-Freeman:


Peter-J-Freeman:
This pull request replaces https://github.com/openvar/variantValidator/pull/877/ and is the final version based on testing during optimisation

PR #880, opened 2026-08-02
Optimise VariantValidator caching and document configuration

Peter-J-Freeman:


Peter-J-Freeman:
Clean commit to replace the pull request https://github.com/openvar/variantValidator/pull/877
The old request is still open as it contains the profiling info. @John-F-Wagstaff.

The code is cleaned, cache settings moved to the settings.py so they can easily be adapted and toggled on and off, including using Environment Variables

VariantValidator/modules/vvMixinInit.py
```diff
@@ -76,6 +127,18 @@ def __init__(self):
...
         """
 
+        # --------------------------------------------------------------
+        # HGVS global configuration
+        # --------------------------------------------------------------
+
+        vvhgvs.global_config.uta.pool_max = 25
+        vvhgvs.global_config.formatting.max_ref_length = 1000000
+
+        if settings.vvHGVS_HDP_CACHE:
```
John-F-Wagstaff, line 137: We probably want a default setting here, or to not set at all (and rely on the vvhgvs default) and only set to 0 when we override it to 0 explicitly, otherwise we just flipped the default from 100 to 0.

Peter-J-Freeman: good point.

Peter-J-Freeman:
@John-F-Wagstaff , can this be merged now?

John-F-Wagstaff (approved):
This looks OK now, we may want to do some re-arranging later, but this works.

PR #882, opened 2026-08-11
Cpr

Peter-J-Freeman:


John-F-Wagstaff:
Please undo the last commit.

As we discussed before I am currently working on splitting the large commits in the 'regex_removal' branch, and doing review changes as I go.  We don't want them to touch at all until that is finished, this will totally break the history or force me to throw away all of the existing work and give up on doing that, if you keep it in.

Peter-J-Freeman:
> Please undo the last commit.

> As we discussed before I am currently working on splitting the large commits in the 'regex_removal' branch, and doing review changes as I go. We don't want them to touch at all until that is finished, this will totally break the history or force me to throw away all of the existing work and give up on doing that, if you keep it in.

OOPS, I jumped the gun on resolving conflicts. Sorry. Now reverted last commit

John-F-Wagstaff:
unfortunately reverting in this way leaves it in the history, I have no idea about how this will interact with my edit's but it is not a clean removal.

It is required and recommended to use revert in projects with more coders, or on public branches that others are using. But this is because doing otherwise will clobber the history, and break any patches authored between the initial commit and the revert. Thankfully this is not the case here. Instead we want to just delete the two last commits and do a force push to reset the HEAD on this branch before them.

(If you have a currently clean version of the branch then the code is `git reset --hard HEAD~1`, twice, once for the revert commit and once for the merge commit that got reverted, followed by a forced push, but I also make a backup branch before doing this kind of thing because '--hard' is exactly what it says and deletes whole commits changes and all)

Peter-J-Freeman:
OK, I'll look into it in the morning and sort it. Apologies again

Peter-J-Freeman:
> (If you have a currently clean version of the branch then the code is git reset --hard HEAD~1, twice, once for the revert commit and once for the merge commit that got reverted, followed by a forced push, but I also make a backup branch before doing this kind of thing because '--hard' is exactly what it says and deletes whole commits changes and all)

Done

